PR #9, opened 2022-06-17
Add release workflow and fix helm linting

terryyylim:
This PR uncomments the release workflow which depends on release.yml file to be found in main branch.

`NONE`

### DIFF
--- a/.github/workflows/xp.yml
+++ b/.github/workflows/xp.yml
@@ -230,7 +230,7 @@ jobs:
       - unit-tests-treatment
       - e2e-tests
       - release-rules
-    uses: gojek/xp/.github/workflows/release.yml@main
+    uses: gojek/turing-experiments/.github/workflows/release.yml@main
     with:
       environment: ${{ needs.release-rules.outputs.release-type == 'dev' && 'manual' || '' }}
     secrets:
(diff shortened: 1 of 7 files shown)


infra/charts/management-service/values.yaml
```diff
@@ -55,7 +55,9 @@ xpManagement:
     class: ""
 
   # -- XP Management Service server configuration.
-  apiConfig: {}
+  apiConfig:
```
krithika369, line 58: Rather than doing this, the helper can check if `sentryConfig` is also set first. See: https://stackoverflow.com/a/68807258

.github/workflows/xp.yml
```diff
@@ -211,27 +211,27 @@ jobs:
...
+        || ( github.event_name != 'pull_request' )
+        || ( github.event.pull_request.head.repo.full_name == github.repository )
+    needs:
+      - lint-python
+      - lint-go
+      - unit-tests-management
+      - unit-tests-treatment
+      - e2e-tests
+      - release-rules
+    uses: gojek/turing-experiments/.github/workflows/release.yml@main
```
krithika369, line 233: Can it not be a relative path?

krithika369 (approved):

.github/workflows/xp.yml
terryyylim: 👍🏻  https://docs.github.com/en/actions/using-workflows/reusing-workflows#calling-a-reusable-workflow

PR #12, opened 2022-06-22
Use alpine base image for Treatment Service

terryyylim:
This PR standardises both Management and Treatment service Docker build uses alpine as its base image. Management Service's publish step is also updated to reuse artifact from previous CI steps.

`NONE`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,0 @@
-# Management Service API
-./management-service/bin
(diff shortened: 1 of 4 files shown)


krithika369:
This LGTM. I guess I missed this in the earlier PR, but can we adopt the same approach for the management service bin too, to reuse the artifact from the previous CI steps? I see we are building it here: https://github.com/gojek/turing-experiments/blob/main/Dockerfile#L10

krithika369 (approved):
Nice! LGTM 🎉

PR #13, opened 2022-06-24
Record no experiment matched as success in the treatment metrics

krithika369:

When no experiment is matched, we are not setting the status as success for the app level histogram metric. This should, in fact, be a success scenario. On the other hand, details on requests with no matched experiment are recorded in the counter `mlp_xp_treatment_service_no_matching_experiment_request_count`.

Fixes #

`NONE`


terryyylim (approved):
LGTM! 🚀

PR #15, opened 2022-07-05
Make Treatment Service max goroutines configurable and add profiling handler

krithika369:
**What this PR does / why we need it**: The Treatment service's liveness probe is currently set to check for a maximum of 100 goroutines exceeding which, the probe will fail. Internally, it has been observed that there are many more concurrently active goroutines than the previous releases. So, this PR does 2 things:
* Adds the default `net/http/pprof` profiling handler at the `/internal/debug/pprof/` endpoint, to get various [diagnostic data](https://go.dev/doc/diagnostics) on demand. It can be invoked as below:
```sh
go tool pprof -http=:8081 http://localhost:8080/v1/internal/debug/pprof/goroutine
```
* Makes the max goroutines value on the liveness probe configurable at deploy time.

With these changes, it should be possible to analyse the runtime performance of the treatment service better and determine if there are any goroutine leaks.

### DIFF
--- a/treatment-service/config/config.go
+++ b/treatment-service/config/config.go
@@ -71,6 +71,7 @@ type SwaggerConfig struct {
 // DeploymentConfig captures the config related to the deployment of Treatment Service
 type DeploymentConfig struct {
 	EnvironmentType string `default:"local"`
+	MaxGoRoutines   int    `default:"100"`
 }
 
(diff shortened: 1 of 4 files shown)


deadlycoconuts (approved):
Looks good to me! Thanks for the changes! 🚀

PR #16, opened 2022-07-07
Add custom segmenters

deadlycoconuts:
This PR ports over the custom segmenters feature that have been introduced to Turing Experiments. Previously this feature was developed internally while the open-sourcing of Turing Experiments was taking place; with this now complete, we're porting over the work that has been completed to our public repository.


.github/workflows/xp.yml
```diff
@@ -52,7 +52,7 @@ jobs:
         with:
           version: v1.41.1
           working-directory: common
-          args: --timeout 3m --verbose
+          args: --timeout 3m --verbose --skip-files=testutils/mocks/ManagementClientInterface.go
```
deadlycoconuts, line 55: I've had to add this because the `typecheck` linter was claiming that the autogenerated `ClientInterface` class uses an unknown method `Called`, which in reality exists as a method of the promoted `mocks.Mock` field of the struct. This is similar to the issue/discussion raised [here](https://github.com/golangci/golangci-lint/discussions/2287) though there doesn't seem to be any conclusive findings/solutions to this issue.

```
testutils/mocks/ManagementClientInterface.go:30:12: _m.Called undefined (type *ClientInterface has no field or method Called) (typecheck)
	ret := _m.Called(_ca...)
```

management-service/models/custom_segmenter.go
```diff
@@ -0,0 +1,678 @@
...
+				allowedValuesSet.Difference(constraintOptionsSet).Len() != 0 {
+				return fmt.Errorf("segmenter values in constraint options do not match those in the allowed values")
+			}
+		}
+	}
+	return nil
+}
+
+// ConvertToTypedValues converts a CustomSegmenter's segmenter values that are untyped to the type specified in
+// the Type field. This method is also used to validate the unknown segmenter value types (i.e. validate values passed
```
krithika369, line 380: > This method is also used to validate the unknown segmenter value types (i.e. validate values passed
// in as user input with respect to the specified type)

Not sure I follow this comment entirely. Do you mean to say that since we will run the type casting on each value, it will be indirectly validating the values?

ui/src/components/config_section/ActivityConfigSection.js
```diff
@@ -1,10 +1,10 @@
+import "components/config_section/ActivityConfigSection.scss";
```
krithika369, line 1: Ooh I think this can still be a relative import as before. `"./ActivityConfigSection.scss"` can be thought of as being local to this component. Do you see any benefit to having the absolute import here?

I have the same comment about other similar changes in this PR. Generally, my rule of thumb is - avoid `../` in imports, at all costs. But, `./` is ok if the imported component can be thought of as a child of the current component or something that goes hand in hand like the style sheets - they will likely live together forever and be refactored together. And, I'd actually argue that the new change makes the import harder to read and picture the relationship to the component.

From a quick search, I didn't find any "best practices" around this online - whether we should adopt absolute imports everywhere or if relative imports are better suited for some scenarios. What are your thoughts?

CC: @leonlnj

ui/src/settings/segmenters/components/form/validation/schema.js
```diff
@@ -0,0 +1,99 @@
...
+
+import { typeOptions } from "settings/segmenters/components/typeOptions";
+
+const nameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/;
+const nameRegexDescription =
+  "Name must begin with an alphanumeric character and have no trailing spaces and can contain letters, numbers, blank spaces and the following symbols: -_()#$%&:.";
+const typeOptionsValues = typeOptions.map((e) => e.value);
+
+// Note: The validation functions below are specifically not defined using the
+// arrow format, for access to `this` from the invocation context:
```
krithika369, line 12: This comment shouldn't be here. Seems like copy-paste error :) We are not defining any special functions in this file.

krithika369 (approved):
Thanks for painstakingly porting over the feature changes and all the refactoring that went in. 🙇🏽‍♀️ (CC: @leonlnj)

As the changes been reviewed in detail in the past, I only glossed over the modifications to each file. I've left some small comments. Please feel free to merge after they are reviewed / addressed.

LGTM. Super excited for this! 🚀

krithika369:
Btw, will we (can we) have a separate PR for the docs as well? https://github.com/gojek/turing-experiments/tree/main/docs

deadlycoconuts:
Oh yeah! I totally forgot about the docs! Oops I'll bring those in with this PR 😅

management-service/models/custom_segmenter.go
deadlycoconuts: Yep that's right; the main purpose of the function is to perform the type cast based on the `type` of the segmenter, but doing that indirectly validates the values too.

ui/src/components/config_section/ActivityConfigSection.js
leonlnj: no opinions, as long as we agree and follow 😆

deadlycoconuts: Okay then, I'll revert `./` changes!

leonlnj (approved):

management-service/models/custom_segmenter.go
krithika369: Got it. Can it be rephrased to include "indirectly validates the type of each value"? I feel that would be clearer.

krithika369:
> Oh yeah! I totally forgot about the docs! Can't remember why I left those out 🤔 Oops I'll bring those in with this PR 😅

Can it be a separate PR though? Just because this one is enormous already. :)

management-service/models/custom_segmenter.go
deadlycoconuts: Okay I've added that phrase in, and now it looks like: 
> ConvertToTypedValues converts a CustomSegmenter's segmenter values that are untyped to the type specified in  the Type field. As this method also indirectly validates the type of each value, it is also used to validate unknown segmenter value types (i.e. validate values passed in as user input with respect to the specified type)

deadlycoconuts:
Ok thanks everyone for look through this! I'll be merging this once the CI pipeline passes! 🚀

PR #18, opened 2022-07-15
Custom Segmenters: Miscellaneous UI Bugfixes

krithika369:
This PR includes small UI fixes.
1. When the experiment data is posted from Create / Edit forms, a recent change (in https://github.com/gojek/turing-experiments/pull/16) was added to parse each segmenter value from string to the registered type. But this is unnecessary as the values in the options list are already of the correct type. The only occasion when it must be parsed from string is when it's keyed into a free-form text area, which happens if there are no pre-defined options for the segmenter (such as with S2ID) - that is handled [here](https://github.com/gojek/turing-experiments/blob/main/ui/src/experiments/components/form/components/segment_config/SegmenterConfigRow.js#L52). Hence, the parsing onSubmit experiment has now been removed in this PR.
Screenshot 2022-07-15 at 12 08 19 PM

2. Small change to "multi-valued" property tooltip for Custom Segmenter.

3. The segmenter type was previously returned in upper case. The PR https://github.com/gojek/turing-experiments/pull/16 also changes this to lower case. But this change hadn't been handled in the UI code fully. Now, removing all the toUpper / toLower conversions and applying the case change only during comparison.

### DIFF
--- a/ui/src/experiments/components/form/validation/schema.js
+++ b/ui/src/experiments/components/form/validation/schema.js
@@ -18,7 +18,7 @@ const experimentTierValues = experimentTiers.map((e) => e.value);
 // arrow format, for access to `this` from the invocation context:
 // https://stackoverflow.com/a/33308151
 
-const validateABTreatmentTraffic = function (items) {
+const validateABTreatmentTraffic = function(items) {
   const sum = items.reduce(
     (total, e) => total + (!!e.traffic ? e.traffic : 0),
     0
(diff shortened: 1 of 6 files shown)


leonlnj (approved):
lgtm

ui/src/experiments/components/form/validation/schema.js
```diff
@@ -219,7 +219,7 @@ const schema = [
                         .typeError("Array elements must all be of type: STRING")
                     );
                   default:
-                    return false;
+                    return yup.array(); // Type is unknown for deactivated segmenters
```
krithika369, line 222: Bugfix: If an experiment has segmenters that have later been removed from the project settings, we won't have the type info in the `segmenterTypes` map. It is incorrect to return `false` here (must be a valid yup schema) and in this circumstance, it is not an error. So we can simply return a `yup.array` without any further type validation.

krithika369:
@leonlnj I snuck in one more small fix. I will merge this MR now but please let me know if you have any comments around that. Thanks!
CC: @deadlycoconuts

PR #19, opened 2022-07-15
Miscellaneous UI bugfixes and refactoring

krithika369:

This PR fixes a handful of small UI bugs.
1. When a segmenter is removed from the Project Settings, and a previously created inactive experiment that uses it is edited, the validation fails. This is addressed in [this change](https://github.com/gojek/turing-experiments/pull/19/files#r923033010).
2. When a 'required' custom segmenter does not have any values in the segment config, the UI would still allow it. This is addressed by [this change](https://github.com/gojek/turing-experiments/pull/19/files#r923035740).
3. Standalone segments were not checking segmenter types. This has been added.

Other than this, some miscellaneous refactoring has been done.

### DIFF
--- a/management-service/services/segmenter_service.go
+++ b/management-service/services/segmenter_service.go
@@ -249,8 +249,10 @@ func (svc *segmenterService) ListSegmenters(
 			}
 			// UpdatedAt and CreatedAt fields are manually updated for custom segmenters but not global segmenters since
 			// global segmenters do not contain these fields
-			formattedCustomSegmenter.UpdatedAt = &customSegmenter.UpdatedAt
-			formattedCustomSegmenter.CreatedAt = &customSegmenter.CreatedAt
+			customSegmenterUpdatedAt := customSegmenter.UpdatedAt
+			customSegmenterCreatedAt := customSegmenter.CreatedAt
+			formattedCustomSegmenter.UpdatedAt = &customSegmenterUpdatedAt
+			formattedCustomSegmenter.CreatedAt = &customSegmenterCreatedAt
 			if params.Status == nil || string(*params.Status) == string(*formattedCustomSegmenter.Status) {
 				allSegmenters = append(allSegmenters, formattedCustomSegmenter)
 			}
--- a/ui/src/experiments/components/form/validation/schema.js
+++ b/ui/src/experiments/components/form/validation/schema.js
@@ -6,6 +6,7 @@ import {
   experimentTiers,
   experimentTypes,
 } from "experiments/components/typeOptions";
+import { segmentConfigSchema } from "segments/components/form/validation/schema";
 
 const nameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/;
 const nameRegexDescription =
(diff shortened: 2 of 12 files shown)


ui/src/experiments/components/form/validation/schema.js
```diff
@@ -6,6 +6,7 @@ import {
   experimentTiers,
   experimentTypes,
 } from "experiments/components/typeOptions";
+import { segmentConfigSchema } from "segments/components/form/validation/schema";
```
krithika369, line 9: Reusing the validation schema defined for the Segments here.

ui/src/services/segmenter/SegmenterStatus.js
```diff
@@ -1,13 +1,13 @@
 export const getSegmenterStatus = (segmenter) => {
   const statusMapping = {
     inactive: {
-      label: "inactive",
```
krithika369, line 4: Changing the case here so that the same map can be used for the status badge in the Segmenter details view as well as in the Segmenters list view, in capital case.

ui/src/services/segmenter/SegmenterScope.js
```diff
@@ -0,0 +1,13 @@
+export const getSegmenterScope = (scope) => {
```
krithika369, line 1: Added this file to match `ui/src/services/segmenter/SegmenterStatus.js`. The definitions will be taken from here instead of inline.

ui/src/segments/components/form/validation/schema.js
```diff
@@ -2,6 +2,62 @@
...
 
 const segmentNameRegex = /^[A-Za-z\d][\w\d \-()#$%&:.]*[\w\d\-()#$%&:.]$/;
+export const segmentConfigSchema = yup.lazy((obj) =>
+  yup.object().shape(
+    Object.keys(obj).reduce((acc, key) => {
+      return {
+        ...acc,
+        [key]: yup.mixed()
+          .when("$segmenterTypes", (segmenterTypes) => {
+            switch ((segmenterTypes[key] || "").toUpperCase()) {
```
krithika369, line 12: Using `|| ""` for when the `type` value is undefined, as with segmenters that are deactivated from the project settings later on.

ui/src/segments/components/form/validation/schema.js
```diff
@@ -2,6 +2,62 @@
...
+                return yup.array(
+                  yup
+                    .string()
+                    .typeError("Array elements must all be of type: STRING")
+                );
+              default:
+                return yup.array(); // Type is unknown for deactivated segmenters
+            }
+          })
+          .when(
```
krithika369, line 44: Previously, the `.when("$segmenterTypes", (segmenterTypes)` check came after this and if so, the chaining of checks will only work if we do `schema.array()` as opposed to a new `yup.array()`. Anyway, I've now moved the type check above instead.

management-service/services/segmenter_service.go
```diff
@@ -249,8 +249,10 @@ func (svc *segmenterService) ListSegmenters(
 			}
 			// UpdatedAt and CreatedAt fields are manually updated for custom segmenters but not global segmenters since
 			// global segmenters do not contain these fields
-			formattedCustomSegmenter.UpdatedAt = &customSegmenter.UpdatedAt
-			formattedCustomSegmenter.CreatedAt = &customSegmenter.CreatedAt
+			customSegmenterUpdatedAt := customSegmenter.UpdatedAt
+			customSegmenterCreatedAt := customSegmenter.CreatedAt
```
deadlycoconuts, line 253: Added these two lines because previously we were assigning the pointer of an iterating `customSegmenter`'s `UpdatedAt` and `CreatedAt` fields to the `customSegmenterUpdatedAt` and `customSegmenterCreatedAt` fields (these are to be returned to as responses of the API call) respectively. This resulted in all the custom segmenters returned to have the same `customSegmenterUpdatedAt` and `customSegmenterCreatedAt` values since they'd all have the same values as the last custom segmenter iterated in the for loop, which was the bug observed by @terryyylim.

image

The solution was to simply initialise new `time.Time` objects at each iteration and then set the  `customSegmenterUpdatedAt` and `customSegmenterCreatedAt` fields with pointers to these objects.

deadlycoconuts (approved):
Looks good to me! 🚀 I've added a minor change in my commit to fix the issue whereby all the timestamps in the list segmenters view are identical.

management-service/services/segmenter_service.go
krithika369: Alternatively, it could be:
```golang
for idx, customSegmenter := range customSegmenters {
    ...
    formattedCustomSegmenter.UpdatedAt = &customSegmenters[idx].UpdatedAt
    ...
}
```

deadlycoconuts: Ah that looks wayyy cleaner 😅

PR #20, opened 2022-07-18
Add Postgres dependency to Management Service helm chart

terryyylim:
Management Service depends on a SQL DB to store experiment configs. Previously, when helm charts were added (https://github.com/gojek/turing-experiments/pull/6), this dependency was not introduced into the Management Service helm chart and this PR addresses that.

This PR also updates helm chart version values to reflect the latest release tag version.

`NONE`


infra/charts/management-service/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "0.0.1"
+appVersion: v0.6.0
```
krithika369, line 2: Besides this, we should keep `xpManagement.image.tag` updated for every release. The `appVersion` here is less relevant. The same applies to the treatment service.

infra/charts/management-service/Chart.yaml
```diff
@@ -1,5 +1,5 @@
...
 description: A Helm chart for Kubernetes Deployment of the XP Managment Service
 name: xp-management
-version: 0.0.1-dev0.1
+version: 0.6.0
```
krithika369, line 5: Since this is the Helm chart version, it need not match the app version (and doesn't really make sense to do so). It can just be `0.1.0`. The same comment applies to the treatment service.

krithika369 (changes requested):

infra/charts/management-service/Chart.yaml
terryyylim: Okay, do we want to publish these changes? If so, perhaps `0.6.1` would make more sense.

infra/charts/management-service/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "0.0.1"
+appVersion: 0.1.0
 description: A Helm chart for Kubernetes Deployment of the XP Managment Service
 name: xp-management
-version: 0.0.1-dev0.1
+version: 0.6.1
```
krithika369, line 5: The other way round, no? Version should be `0.1.0`

infra/charts/management-service/values.yaml
```diff
@@ -1,11 +1,11 @@
...
   image:
     # -- Docker registry for XP Management Service image
-    registry: docker.io/
+    registry: ghcr.io
     # -- Docker image repository for XP Management Service
-    repository: xp-management
+    repository: gojek/turing-experiments/xp-management
     # -- Docker image tag for XP Management Service
-    tag: latest
+    tag: v0.6.1
```
krithika369: Why `v0.6.1` as opposed to `v0.6.0` ? There is real change to the `v0.6.0` release since (all the new changes are for CS).

terryyylim: Yeah, noticed that, updated to `v0.6.0`.

infra/charts/management-service/Chart.yaml
krithika369: We only need to publish the chart for this PR - no changes to the app itself, correct? So `0.6.0` is fine.

PR #21, opened 2022-07-19
Bump up chart values and version for v0.7.0

krithika369:

Updating the Helm chart values in preparation for app release `v0.7.0`. README has been autogenerated using `helm-docs`.

### DIFF
--- a/infra/charts/management-service/Chart.yaml
+++ b/infra/charts/management-service/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: v0.6.0
+appVersion: v0.7.0
 description: A Helm chart for Kubernetes Deployment of the XP Managment Service
 name: xp-management
-version: 0.1.0
+version: 0.1.1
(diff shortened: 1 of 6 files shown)


infra/charts/management-service/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: v0.6.0
+appVersion: v0.7.0
 description: A Helm chart for Kubernetes Deployment of the XP Managment Service
 name: xp-management
-version: 0.1.0
+version: 0.1.1
```
deadlycoconuts, line 5: Just wondering, are we just bumping up patch version numbers for the chart versions for each new release?

deadlycoconuts (approved):
Looks good to me! Just left a tiny comment but otherwise all's good!

infra/charts/management-service/Chart.yaml
krithika369: The chart itself has no change - it's just updating the tag value in `values.yaml` with the new app version. That's why.

PR #22, opened 2022-07-20
[BUG] Fix experiment search filters when using real segmenter values

deadlycoconuts:
This PR introduces a bug fix that occurs when filtering experiments using segmenter values - when using segmenter values that are of the type real (`float` in go), experiments that have these values specified do not get filtered and displayed correctly in the `ListExperimentView`. 

E.g. Searching for experiments with the segmenter value corresponding to 1.4 (`o4`) for `custom_float` does not return the experiment named 'test':

Screenshot 2022-07-20 at 6 45 56 AM
Screenshot 2022-07-20 at 6 42 20 AM
Screenshot 2022-07-20 at 6 43 27 AM

**Details**
The bug is caused by the backend storing float values as strings with trailing zeroes e.g. "1.40000", leading the gorm string condition search queries to fail when given search parameters without them i.e. "1.4". 

Screenshot 2022-07-20 at 6 12 22 AM

The simple fix proposed in this PR thus enforces the strings generated from float values to have no trailing zeroes, by using the `strconv.FormatFloat` method detailed [here](https://pkg.go.dev/strconv#FormatFloat).

Although it is sufficient to apply this change to the storing of segmenter values used in experiments (i.e. in `management-service/models/experiment_segment.go`), a similar change has been made for the storing of segmenter values in custom segmenters for additional consistency.

**Extras**
This PR also includes a tiny code refactor from the comment made in https://github.com/gojek/turing-experiments/pull/19#discussion_r923221867.

### DIFF
--- a/management-service/services/segmenter_service.go
+++ b/management-service/services/segmenter_service.go
@@ -235,7 +235,7 @@ func (svc *segmenterService) ListSegmenters(
 		if err != nil {
 			return nil, err
 		}
-		for _, customSegmenter := range customSegmenters {
+		for idx, customSegmenter := range customSegmenters {
 			if err := customSegmenter.FromStorageSchema(segmenterTypes); err != nil {
 				return nil, err
 			}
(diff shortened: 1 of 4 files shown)


management-service/services/segmenter_service.go
```diff
@@ -249,10 +249,8 @@ func (svc *segmenterService) ListSegmenters(
 			}
 			// UpdatedAt and CreatedAt fields are manually updated for custom segmenters but not global segmenters since
 			// global segmenters do not contain these fields
-			customSegmenterUpdatedAt := customSegmenter.UpdatedAt
-			customSegmenterCreatedAt := customSegmenter.CreatedAt
-			formattedCustomSegmenter.UpdatedAt = &customSegmenterUpdatedAt
-			formattedCustomSegmenter.CreatedAt = &customSegmenterCreatedAt
+			formattedCustomSegmenter.UpdatedAt = &customSegmenters[idx].UpdatedAt
+			formattedCustomSegmenter.CreatedAt = &customSegmenters[idx].CreatedAt
```
deadlycoconuts, line 253: Slipped this in in response to https://github.com/gojek/turing-experiments/pull/19#discussion_r923221867.

krithika369 (approved):
Thanks for the bug fix and the detailed explanation. LGTM 🚀

PR #23, opened 2022-07-21
Add Turing Experiment Engine XP Plugin

terryyylim:
This PR adds the following:

1. Auto-conversion of Segmenter value from request payload from string to respective Segmenter type via `cast` library. This is required since Segmenter type information is most suited to be left in Segmenter Runner's Transform method, rather than the Experiment Engine plugin itself.
2. Experiment Engine plugin for [Turing](https://github.com/gojek/turing).
3. CICD build/publish step to expose XP Experiment Engine plugin for first-class support in Turing.

`NONE`


plugins/turing/runner/experiment_runner.go
```diff
@@ -0,0 +1,151 @@
...
+	body []byte,
+	options runner.GetTreatmentOptions,
+) (*runner.Treatment, error) {
+	logger := log.With("turing_req_id", options.TuringRequestID)
+
+	// Get the request parameters for the current request
+	requestParams := er.getRequestParams(logger, reqHeader, body)
+
+	// Create Request Payload
+	reqPayload, err := json.Marshal(requestParams)
```
terryyylim, line 48: Marshal `map[string]string` instead of `map[string]interface{}` since Experiment Engine plugin should not know about Segmenter type conversions.

plugins/turing/main.go
```diff
@@ -0,0 +1,18 @@
+package main
```
terryyylim, line 1: Refactored Experiment Engine plugin folder structure:
- Introduced `config`, `manager` and `runner` packages
- Removed `cmd` package

plugins/turing/Makefile
```diff
@@ -0,0 +1,46 @@
...
+# ==================================
+
+.PHONY: build
+build:
+	@echo "Building binary..."
+	go build -o ./bin/plugin
+
+.PHONY: build-image
+build-image: vendor version
+	@$(eval IMAGE_TAG = $(if $(DOCKER_REGISTRY),$(DOCKER_REGISTRY)/,)xp-engine-plugin:${VERSION})
```
terryyylim, line 33: `xp-engine-plugin` to be published to ghcr - https://github.com/gojek/turing-experiments/runs/7443640421?check_suite_focus=true (successful run)

.github/workflows/plugins.yml
```diff
@@ -0,0 +1,115 @@
...
+
+  publish-turing-experiment-engine-xp-plugin:
+    # Automatically publish release and pre-release artifacts.
+    #
+    # As for dev releases, make it possible to publish artifacts
+    # manually by approving 'deployment' in the 'manual' environment.
+    #
+    # Dev build can be released either from the 'main' branch or
+    # by running this workflow manually with `workflow_dispatch` event.
+#    if: >-
```
terryyylim, line 76: Currently commented to trigger workflow via PR. Will be uncommented once PR is approved prior to merging.

.github/workflows/plugins.yml
```diff
@@ -0,0 +1,115 @@
...
+        working-directory: plugins/turing
+        env:
+          DOCKER_REGISTRY: ghcr.io/${{ github.repository }}
+          PLUGIN_NAME: xp
+        run: |
+          set -o pipefail
+          make build-image | tee output.log
+          echo "::set-output name=xp-engine-plugin-image::$(sed -n 's%Building docker image: \(.*\)%\1%p' output.log)"
+
+#      - name: Publish Turing Experiment Engine XP Plugin Docker Image
```
terryyylim, line 114: Currently commented to trigger workflow via PR. Will be uncommented once PR is approved prior to merging.

common/utils/typeconverter.go
```diff
@@ -71,9 +75,74 @@ func SegmenterValueToInterface(value *_segmenters.SegmenterValue) interface{} {
 	}
 }
 
-func InterfaceToSegmenterValue(value interface{}, valueType *_segmenters.SegmenterValueType) *_segmenters.SegmenterValue {
+func GetIntSegmenter(value interface{}, key string, segmenter string) (*int64, error) {
```
terryyylim, line 78: `GetIntSegmenter`, `GetFloatSegmenter`, `GetBoolSegmenter` adds support for casting string request values to appropriate segmenter type values.

.github/workflows/plugins.yml
```diff
@@ -0,0 +1,115 @@
+name: plugins
+
+on:
+  # Automatically run CI on Release and Pre-Release tags and main branch
+  # (only if there are changes to relevant paths)
+  push:
+    tags:
+      - "xp-engine-plugin/v[0-9]+.[0-9]+.[0-9]+*"
```
krithika369, line 8: Nit: Can it just be `turing-plugin` for brevity in the tagging? Since it belongs to the `turing-experiments` repository, it would be clear enough that it's for XP, no? Also the job names: `turing-experiment-engine-xp-plugin-tests` -> `turing-plugin-tests`.

common/utils/typeconverter.go
krithika369: Cool. Although now, the implementation looks pretty bulky. Shall we instead use [this library](https://github.com/spf13/cast)? It seems to cater to the corner cases too. If using the library, may be we don't need the utility methods anymore and can be directly called where required.

plugins/turing/main.go
krithika369: Why is the `cmd` package removed? It's still useful for development.

plugins/turing/runner/experiment_runner_test.go
```diff
@@ -0,0 +1,136 @@
+package runner
```
krithika369, line 1: I notice the integration test has not been ported over (which makes sense because we're not running the treatment service for this). Can we instead add a similar test into this file using appropriate mocking?

krithika369 (changes requested):
Thanks for this PR! Left some comments. Will have another quick look once they are addressed.

common/utils/typeconverter.go
terryyylim: I think the cast  package is definitely useful, but it does catch some corner cases which are not what we want. 

**Examples**
```
cast.ToFloat(true)            // 1
cast.ToFloat(false)          // 0
cast.ToInt(true)               // 1
cast.ToInt(false)              // 0
```
For the above cases, our current implementation raise errors for us, which is what we want, rather than casting to binary values.

krithika369: Fair enough. But:
* Should we really worry about the user registering an int type and sending in a boolean value?
* Can this be considered a feature of XP rather than a bug? 😅 We can argue that `True: 1` and `False: 0` are conventional?

terryyylim: >  Should we really worry about the user registering an int type and sending in a boolean value?

I'd say its user problem, just that we won't be providing any informative feedback loop to the user immediately, they'd likely have to debug themselves when they notice treatments are not fetched correctly.

> Can this be considered a feature of XP rather than a bug? 😅 We can argue that True: 1 and False: 0 are conventional?

I guess we could, I see it as a tradeoff between ease of debugging vs. maintenance, and in this case, it does feel more like a user problem than something we should be catching 🤔 

We'll go with `cast` then?

krithika369: Yes, let's reduce our maintainable code considering the problem is an atypical scenario.

terryyylim:
> Why is the cmd package removed? It's still useful for development.

Could you elaborate a little more on what you mean by `useful for development`?

krithika369 (approved):
Thanks for the changes! Left a comment on the integration test case to use mocking instead of patching and some other small comments. Once you've reviewed / addressed them, please feel free to merge. 🚀

PR #25, opened 2022-07-25
Add system architecture info

krithika369:

This PR captures the system architecture into the README.


terryyylim (approved):
Thanks for this @krithika369 ! Just a nit - Could we have the `Configure Experiments` and `Fetch treatment requests from clients` both come from a "smiley face" icon for consistency?

krithika369:
> Thanks for this @krithika369 ! Just a nit - Could we have the `Configure Experiments` and `Fetch treatment requests from clients` both come from a "smiley face" icon for consistency?

The fetch treatment call likely comes from a different system (Turing routers / backend) which is why I left it out. We can refine it as we go along.

Thanks for the quick review! Will merge once the CI completes.

PR #26, opened 2022-07-26
Refactor Treatment Service Integration test Kafka setup

terryyylim:
Treatment Service's integration test started becoming flaky after Go version was updated in this [PR](https://github.com/gojek/turing-experiments/pull/24).

This issue has surfaced before on our internal Gitlab runners, and our interim fix was to run the following.
```
docker container prune
docker network prune
docker volume prune
```

According to Github Actions runner [documentation](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#using-a-github-hosted-runner), when the job has finished, the VM is automatically decommissioned. Thus, for each new trigger of re-running a failed job, there should not be any unused network etc. that might cause issues.

For our current [implementation](https://github.com/gojek/turing-experiments/blob/main/treatment-service/internal/testutils/kafka.go), the hypothesis here is that when integration tests run, the docker objects which have been created are not cleaned up properly. This is tricky to test considering we use github-hosted runners and we don't have access to investigate the runner VM itself. What is evident is the lack of cleanup utility in the [file](https://github.com/gojek/turing-experiments/blob/main/treatment-service/internal/testutils/kafka.go) nor the [teardown function](https://github.com/gojek/turing-experiments/blob/main/treatment-service/integration-test/fetch_treatment_it_test.go#L402).

Considering the following points which adds complexity to maintainability, I decided to refactor how `test-containers` are setup:
- Current implementation uses a relative older Kafka version which requires Zookeeper, and a docker network for communication between the 2 components
- Convoluted implementation in [testutils](https://github.com/gojek/turing-experiments/blob/main/treatment-service/internal/testutils/kafka.go)

This PR:
1. Refactors the integration test Kafka dependency's setup, by using the latest Kafka version which uses Kraft that was introduced to remove Kafka's dependency on ZooKeeper for metadata management. This should ideally make the test suite easier to maintain/debug.
2. Updates default buffered channel size for Treatment logs to a smaller number

`Fix flake Treatment Service integration test`


treatment-service/integration-test/fetch_treatment_it_test.go
```diff
@@ -367,14 +367,17 @@ func (suite *TreatmentServiceTestSuite) SetupSuite() {
 	suite.managementServiceClient = managementClient
 	suite.managementServiceServer = managementServer
 
-	kafkaCluster := tu.NewKafkaCluster()
-	err = kafkaCluster.StartCluster()
+	composeFilePaths := []string{"docker-compose.yaml"}
+	kafka := testcontainers.NewLocalDockerCompose(composeFilePaths, strings.ToLower(uuid.New().String()))
```
krithika369, line 371: Can we explicitly specify the service name that should be started up - `kafka` in this case (something to the effect of `docker-compose up -d kafka`)? This is so that in case more services are added to the compose file, we can limit what should be started for each test. Or perhaps we can restructure the compose file to `docker-compose/kafka.yaml`. wdyt?

krithika369:
Thanks for the refactoring - it LGTM on the whole. But I have some questions about why you describe in the PR that this is a "hypothesis":
* Were you able to repro the test failure locally / on the runner?
* If so, did you manage to figure out where the failure/timeout was happening - with some debug logs or otherwise ?

terryyylim:
> Thanks for the refactoring - it LGTM on the whole. But I have some questions about why you describe in the PR that this is a "hypothesis":
> 
> * Were you able to repro the test failure locally / on the runner?
> * If so, did you manage to figure out where the failure/timeout was happening - with some debug logs or otherwise ?

- I was able to repro failing tests when the network was not cleaned up (locally), but not on the runner
- The failure typically occurs when the connection between Kafka and Zookeeper was not established correctly and communication cannot happen
- Hence, I "linked" the issue to what happened on our internal Gitlab CICD previously

krithika369:
> * The failure typically occurs when the connection between Kafka and Zookeeper was not established correctly and communication cannot happen

Ok. Thanks for the clarification. The fix makes sense to me. Let's keep an eye out and revisit this if the issue comes up again.

krithika369 (approved):

terryyylim:
> > * The failure typically occurs when the connection between Kafka and Zookeeper was not established correctly and communication cannot happen
> 
> Ok. Thanks for the clarification. The fix makes sense to me. Let's keep an eye out and revisit this if the issue comes up again.

Seems like it's still failing: https://github.com/gojek/turing-experiments/runs/7513741149?check_suite_focus=true. Might be some other issue then 🤔 I'll try to investigate further.

treatment-service/config/config.go
```diff
@@ -37,7 +37,7 @@ type Config struct {
 
 type AssignedTreatmentLoggerConfig struct {
 	Kind                 AssignedTreatmentLoggerKind `default:""`
-	QueueLength          int                         `default:"1073741824"`
```
terryyylim, line 40: Not sure why the default value is `1073741824` (which coincidentally coincides with the size of 1 gibibyte), but basically `QueueLength` is used for creating a [buffered channel](https://github.com/gojek/turing-experiments/blob/main/treatment-service/monitoring/treatmentlogger.go#L99) to receive and publish `AssignedTreatmentLog`.

By adding more logs to debug the issue, these failed tests pin-point the failing line to where the buffered channel is created.

`Prior to using a lower channel size`
https://github.com/gojek/turing-experiments/runs/7515128336?check_suite_focus=true
https://github.com/gojek/turing-experiments/runs/7514891584?check_suite_focus=true
https://github.com/gojek/turing-experiments/runs/7515034880?check_suite_focus=true

I dug deeper into setting a high value for a buffered channel and this seems to create a problem where too much memory is allocated. One thing that I haven't fully figured out though, is why we didn't face issues previously when `makechan` is supposed to throw an [error](https://github.com/golang/go/blob/master/src/runtime/chan.go#L76), i.e `1<<16 = 65536` and `1073741824` clearly exceeds the limit.

treatment-service/integration-test/docker-compose/kafka.yaml
```diff
@@ -0,0 +1,30 @@
+version: '3'
```
terryyylim, line 1: Confluent Kafka Docker-compose setup copied from https://github.com/confluentinc/cp-all-in-one/blob/7.0.1-post/cp-all-in-one-kraft/docker-compose.yml.

treatment-service/integration-test/docker-compose/kafka.yaml
```diff
@@ -0,0 +1,30 @@
...
+      KAFKA_PROCESS_ROLES: 'broker,controller'
+      KAFKA_NODE_ID: 1
+      KAFKA_CONTROLLER_QUORUM_VOTERS: '1@broker:29093'
+      KAFKA_LISTENERS: 'PLAINTEXT://broker:29092,CONTROLLER://broker:29093,PLAINTEXT_HOST://0.0.0.0:9092'
+      KAFKA_INTER_BROKER_LISTENER_NAME: 'PLAINTEXT'
+      KAFKA_CONTROLLER_LISTENER_NAMES: 'CONTROLLER'
+      KAFKA_LOG_DIRS: '/tmp/kraft-combined-logs'
+    volumes:
+      - ./update_run.sh:/tmp/update_run.sh
+    command: "bash -c 'if [ ! -f /tmp/update_run.sh ]; then echo \"ERROR: Did you forget the update_run.sh file that came with this docker-compose.yml file?\" && exit 1 ; else /tmp/update_run.sh && /etc/confluent/docker/run ; fi'"
```
krithika369, line 30: Nit: Since we have the `update_run.sh` now, may be cleaner to move to  nested `kafka` directory:
* treatment-service/integration-test/docker-compose/kafka/docker-compose.yaml
* treatment-service/integration-test/docker-compose/kafka/update_run_sh.yaml

treatment-service/config/config.go
terryyylim: Ah, it was progressive logging.

So originally, the logs only showed `Detected KafkaLogger...`, then it terminated with what was shown in the screenshot. I added another [log previously](https://github.com/gojek/turing-experiments/pull/26/commits/b6acfd3d70c906782f1b873958ace5ebe8573920#diff-b9a96c0bbd14362b23737368d794bc95f998578dd7a7574c7e54471ed0000ae5L102), which should have shown `Configuring publisher...` should the channel be correctly created, but that logging statement was never shown.

image

treatment-service/integration-test/docker-compose/kafka.yaml
krithika369: Would be nice to add this as inline comment in the code so it's easy to identify in the future.

treatment-service/config/config.go
krithika369: I see.. Ok. Also, shall we trigger multiple test runs (like 5-10 times) to see if there is any more flakiness? If not, it could be safe to assume the buffer size change fixed it and we can merge this PR.

terryyylim: 👍🏻 Agreed, I'll trigger it a few more times on this PR prior to merging.

PR #27, opened 2022-07-27
Update helm chart to check for nil pointer

terryyylim:
While integrating XP helm chart with Turing, nil pointer [error](https://github.com/gojek/turing/runs/7535483563?check_suite_focus=true) was raised when linting the helm chart. This PR addresses that.

`NONE`

### DIFF
--- a/infra/charts/management-service/templates/_helpers.tpl
+++ b/infra/charts/management-service/templates/_helpers.tpl
@@ -39,7 +39,7 @@ Common labels
 */}}
 {{- define "xp.labels" -}}
 helm.sh/chart: {{ include "xp.chart" . }}
-{{- with .Values.xpManagement.extraLabels }}
+{{- with (.Values.xpManagement).extraLabels }}
 {{- toYaml . | nindent 0 }}
 {{- end }}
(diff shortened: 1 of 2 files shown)


infra/charts/management-service/templates/_helpers.tpl
```diff
@@ -39,7 +39,7 @@ Common labels
 */}}
 {{- define "xp.labels" -}}
 helm.sh/chart: {{ include "xp.chart" . }}
-{{- with .Values.xpManagement.extraLabels }}
+{{- with (.Values.xpManagement).extraLabels }}
```
krithika369, line 42: Has this been tested? I'm not sure if we need it to be `((.Values.xpManagement).extraLabels)` (additional paranthesis) instead.

krithika369 (approved):
Left one small comment, please feel free to merge if that's been reviewed.

infra/charts/management-service/templates/_helpers.tpl
terryyylim: Yep, tested this via Turing's helm chart. No need for the additional parenthesis.

PR #29, opened 2022-08-15
Replace old module path gojek/xp with caraml-dev/xp

deadlycoconuts:
In line with the migration of the Turing Experiments repository from the [gojek organisation](https://github.com/gojek) (`gojek/turing-experiments`) to the new [CaraML organisation](https://github.com/caraml-dev) (`caraml-dev/xp`), we need to update certain dependencies that were pointing to the original repo (e.g. Docker image repositories, Go dependencies pointing to `gojek/turing`, etc.). In addition, we are also updating the Go module paths from `github.com/gojek/xp` to `github.com/caraml-dev/xp` to reflect the new changes.

None


ui/public/index.html
```diff
@@ -6,7 +6,7 @@
   <link rel="icon" href="%PUBLIC_URL%/favicon.ico" />
   <meta name="viewport" content="width=device-width, initial-scale=1" />
   <meta name="theme-color" content="#000000" />
-  <meta name="description" content="Machine Learning Platform" />
+  <meta name="description" content="CaraML" />
```
deadlycoconuts, line 9: Updated this to make the migration consistent with the future changes to the MLP UI i.e. renaming of 'Machine Learning Platform' to 'CaraML'.

terryyylim (commented):
LGTM! Thanks @deadlycoconuts . But I'm wondering if we should take this chance to shift from `turing-experiments` to `xp`? wdyt @krithika369?

krithika369:
> LGTM! Thanks @deadlycoconuts . But I'm wondering if we should take this chance to shift from `turing-experiments` to `xp`? wdyt @krithika369?

Yes! Let's do it. :)

deadlycoconuts:
Just to confirm, I'm assuming both of you are referring to a repo name change right? 😅

README.md
```diff
@@ -65,10 +65,10 @@ To test authorization for Management Service locally, make the following changes
 To use the XP Go modules as API dependencies, `replace` directives need to be used when consuming both the Management and Treatment API modules since the API modules uses local relative path.
 
 ```go
-replace github.com/gojek/xp/clients => github.com/gojek/turing-experiments/clients v0.0.0
-replace github.com/gojek/xp/common => github.com/gojek/turing-experiments/common v0.0.0
-replace github.com/gojek/xp/management-service => github.com/gojek/turing-experiments/management-service v0.0.0
-replace github.com/gojek/xp/treatment-service => github.com/gojek/turing-experiments/treatment-service v0.0.0
+replace github.com/caraml-dev/xp/clients => github.com/caraml-dev/xp/clients v0.0.0
```
terryyylim, line 68: These replace directives are incorrect.

If we're keeping the use of `gojek/xp` references when using the Go modules as API dependencies, then the following should be the right directives:
```
replace github.com/gojek/xp/clients => github.com/gojek/turing-experiments/clients v0.0.0
...

Considering the shift to `caraml-dev` organisation is a permanent one, all internal repositories that reference Github XP could have their dependencies changed to point directly to `caraml-dev/xp` as well.
```

docs/how-to/07_running_experiments.md
```diff
@@ -14,7 +14,7 @@ all the variable(s) must be provided in the fetch treatment request body. For ea
 
 ## Running Experiments with Turing Routers
 
-When deploying a Turing router, the experiment engine can be configured to 'Turing Experiments'. For more information, check [Turing - Creating a Router](https://github.com/gojek/turing/tree/main/docs/how-to/create-a-router).
+When deploying a Turing router, the experiment engine can be configured to 'Turing Experiments'. For more information, check [Turing - Creating a Router](https://github.com/caraml-dev/turing/tree/main/docs/how-to/create-a-router).
```
terryyylim, line 17: Should the URL point to docs repository now that we have a consolidated one?

terryyylim (approved):
Left some minor comments, but overall LGTM, thanks for this!

README.md
deadlycoconuts: > If we're keeping the use of gojek/xp references when using the Go modules as API dependencies
I think we're changing the name of the Go modules from `gojek/xp` to `caraml-dev/xp`, or at least that's what we're doing for Turing too.

> Considering the shift to caraml-dev organisation is a permanent one, all internal repositories that reference Github XP could have their dependencies changed to point directly to caraml-dev/xp as well.
Yep, I'll update those once this has been merged :)

docs/how-to/07_running_experiments.md
deadlycoconuts: Hmmm, like the CaraML docs repository? Hmm we can do that but it wouldn't bring the user to the Turing repo's URL though, which I believe would be more helpful if the user wishes to explore what Turing is after clicking on the link?

README.md
deadlycoconuts: I've just removed the replace directives from the README since they aren't useful anymore (the Go module paths are now named exactly as their URLs)

PR #30, opened 2022-08-15
Satisfy Istio's requirement for named ports  and update pre-commit hooks

terryyylim:
Istio [requires](https://istio.io/v1.0/docs/setup/kubernetes/spec-requirements/) service ports to be named in order to work correctly with the routing features. (It is also explained here in detail: [Protocol Selection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/)). While the current chart works correctly with some basic usage of Istio (i.e. exposing a service via VirtualService), it fails to work as expected inside Istio's Service mesh.

This MR adds http name to XP's Services and bumps up a patch version of the chart, similar to how Turing's helm chart has been [updated](https://github.com/caraml-dev/turing/pull/223).

In addition, the [pre-commit](https://pre-commit.com/#intro) tool has been integrated which helps automate some additional checks / actions when some files are being modified (eg: regenerating the Helm docs with changes to the chart values). The  `.pre-commit-config.yaml` file has each of these hooks:
* Helm docs generation
* Linting of Golang, JS code (local hook, using the respective `make` target)

`NONE`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,5 +1,3 @@
-default_stages:
-  - push
 repos:
   - repo: https://github.com/norwoodj/helm-docs
     rev: v1.11.0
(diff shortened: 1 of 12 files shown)


krithika369 (commented):
We don't need to make this change for the treatment service, per the original motivation (letting Istio handle the authentication), although I guess there is no harm in doing so. But. a couple of other comments:
* The helm docs must be regenerated. Could you also integrate the changes from the recent Turing MR to do this automatically on commit? https://github.com/caraml-dev/turing/pull/226
* Could you link the related Turing MR in the PR description, for better context? https://github.com/caraml-dev/turing/pull/223

krithika369 (approved):
LGTM. Thanks for this PR!

PR #34, opened 2022-09-12
Allow skipping experiment variables in Treatment Request

krithika369:

The Fetch Treatment call to the Treatment Service would validate that all the experiment variables (for the project's chosen segmenters as well as the randomization key) were present in the request body, and if any of them were missing, the request would fail immediately. However, this is not ideal for some scenarios:
* User has only configured Switchback experiments (here, the randomization key is not needed)
* User has configured experiments with some / all segmenters being optional (here, we could attempt to match experiments with optional segmenter where the experiment variables are missing for a certain segmenter).

This PR makes it possible to skip one or more experiment variables in the fetch treatment request. Experiment matches will be carried out using the available request parameters and the randomization key will only be validated if an A/B experiment is matched. In order to support this, the treatment service has been updated to remove / defer some of the validations. Accordingly, the XP micro frontend / plugin for Turing are also updated such that a `None` field source is introduced. This will be used if the user wants to skip configuring certain experiment variables.

### Treatment Service
* `treatment-service/controller/treatment.go` - Remove schema validation in the handler for the fetch treatment call
* `treatment-service/services/segmenter_service.go` - Allow `GetTransformation` to return an empty segmenter values list (in which case, we will only match those experiments defining that segmenter to be optional)
* `treatment-service/services/schema_service.go` - Clean up unused validations

### Turing API Plugin
* `plugins/turing/config/config.go` - Refactor configs. Please refer to the note in the PR description for details of impact.
* `plugins/turing/manager/experiment_manager.go` - Update yup validation schema to handle `none` as field source and missing field value
* `plugins/turing/runner/experiment_runner.go` - Handle missing field info

### Turing Microfrontend UI
* `ui/src/turing/components/form/variables_config/components/FieldSourceFormLabel.js` - Add `none` field source
* `ui/src/turing/components/form/variables_config/VariablesConfigPanel.js` - Make `none` the default field source
* `ui/src/turing/components/form/variables_config/VariableConfigRow.js` - If field source is `none`, clear the field name

Screenshot 2022-09-12 at 12 48 40 PM

Screenshot 2022-09-12 at 12 48 52 PM

### Other Changes
* Added lint step for the Turing XP API plugin
* Updated the `PB_VERSION` used for proto compilation

## Note
Where the XP plugin version is saved to the router details (in the respective internal DB), due to the change detailed in [this comment](https://github.com/caraml-dev/xp/pull/34/files#r967984748).


plugins/turing/Makefile
```diff
@@ -16,9 +16,12 @@ vendor:
 # Code dependencies recipes
 # ==================================
 
-tidy-plugin:
+tidy:
 	go mod tidy
 
+lint:
+	golangci-lint run
```
krithika369, line 23: Add lint target

plugins/turing/config/config.go
```diff
@@ -1,22 +1,23 @@
...
+type FieldSource string
+
+const (
+	// PayloadFieldSource is used to represent the request payload
+	PayloadFieldSource FieldSource = "payload"
+	// HeaderFieldSource is used to represent the request header
+	HeaderFieldSource FieldSource = "header"
+	// NoneFieldSource is used to represent that there is no field source,
+	// i.e., the variable has not been configured.
+	NoneFieldSource FieldSource = "none"
```
krithika369, line 19: Adding `NoneFieldSource` for the scenario where the user wants to skip configuring a variable.

Makefile
```diff
@@ -15,7 +15,7 @@ ifeq ($(UNAME_S),Darwin)
 endif
 
 PB_URL="https://github.com/protocolbuffers/protobuf/releases"
-PB_VERSION=3.15.8
+PB_VERSION=3.19.4
```
krithika369, line 18: Bumping up this version because it was used in `common/segmenters/segmenters.pb.go` in https://github.com/caraml-dev/xp/pull/16

plugins/turing/manager/experiment_manager.go
```diff
@@ -29,6 +29,9 @@ func init() {
 // Default timeout for requests made to the XP API server
 const defaultRequestTimeout = time.Second * 5
 
+// TODO: Move the validation within the microfrontend component instead of exposing the
```
krithika369, line 32: We have an internal card tracking this.

plugins/turing/config/config.go
```diff
@@ -1,22 +1,23 @@
...
+	"fmt"
 	"sync"
 
-	"github.com/caraml-dev/turing/engines/experiment/pkg/request"
 	"github.com/go-playground/validator/v10"
 )
 
-// RequestParameter captures a single parameter's config for parsing the incoming
-// request and extracting values
-type RequestParameter struct {
```
krithika369, line 12: This struct is almost exactly the same as `type Variable` below. So, replaced usage with that instead.

But, this change will affect the parameter names set by the experiment manager / used by the experiment runner (field `parameter` -> `name`). Therefore, we would need to update the plugin version to latest, for all existing records in the DB, to support redeploy of existing routers with new experiment manager.

treatment-service/controller/treatment.go
```diff
@@ -126,19 +126,6 @@ func (t TreatmentController) FetchTreatment(w http.ResponseWriter, r *http.Reque
 		return
 	}
 
-	err = t.SchemaService.ValidateSchema(projectId, filterParams.AdditionalProperties)
```
krithika369, line 129: No longer needed here, checks are already present further down the request processing funnel.

treatment-service/models/storage.go
```diff
@@ -201,13 +202,14 @@ func (i *ExperimentIndex) matchRealSetSegment(segmentName string, value float64)
 }
 
 func (i *ExperimentIndex) matchSegment(segmentName string, values []*_segmenters.SegmenterValue) Match {
-	_, exists := i.Experiment.Segments[segmentName]
-	if !exists {
-		// Optional segmenter, first element from segmenter values (cannot be empty)
-		return Match{Strength: MatchStrengthWeak, Value: values[0]}
+	if len(values) == 0 {
```
krithika369, line 205: We only need to have this check for optional segmenter here for `len(values) == 0` - the rest is handled within the check for each type of value (eg: `matchFlagSetSegment` or `matchStringSetSegment`)

deadlycoconuts (approved):
LGTM, just left 1 tiny comment. Thanks a lot for working on this! 🚀

terryyylim (approved):
LGTM, thanks for PR! 🚀  @krithika369

plugins/turing/Makefile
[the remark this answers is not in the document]
krithika369: Added the log. As for the timeout, the default value is `1m` and since the plugin codebase is small, it should suffice.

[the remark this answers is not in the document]
deadlycoconuts: Thanks for the changes!

krithika369:
Thanks, both, for the prompt review! I'll be merging this PR once the pipeline succeeds.

PR #35, opened 2022-09-15
Bugfix: Cater to optional segmenter matches in filterByLookupOrder

krithika369:
When multiple experiments are matched by the Fetch Treatment API, the hierarchical experiment filters ensure that ultimately, only one experiment will be selected. However, in a case where there are multiple experiments after `filterByMatchStrength` and there are missing segmenter values to the Fetch Treatment request, the `filterByLookupOrder` function would simply eliminate the experiment(s) from the matches. But, in fact, the overall matches should only be updated if there are non-zero matches for a given segmenter value (which will automatically handle `nil` segmenter values).

This PR addresses the issue above by only updating the filtered experiments if the matches in the current iteration is non-zero, much like the other filters for hierarchical experiments.

### DIFF
--- a/treatment-service/services/experiment_service.go
+++ b/treatment-service/services/experiment_service.go
@@ -124,7 +124,6 @@ func (es *experimentService) filterByLookupOrder(
 	segmenters []string,
 	segmenterTypes map[string]schema.SegmenterType,
 ) []*models.ExperimentMatch {
-	// Stop search when we have at least 1 match
 	filtered := matches
 	for _, segmenter := range segmenters {
 		orderedValues := filters[segmenter]
@@ -155,7 +154,8 @@ func (es *experimentService) filterByLookupOrder(
 				}
 			}
 		}
-		// currentFilteredList could be 0 in case of weak matches
+		// currentFilteredList could be 0 in case of weak matches, where the segmenter info was
+		// missing in the original request (i.e., where orderedValues is empty for a given segmenter).
 		if len(currentFilteredList) > 0 {
 			filtered = currentFilteredList
 		}
--- a/treatment-service/services/experiment_service.go
+++ b/treatment-service/services/experiment_service.go
@@ -128,38 +128,37 @@ func (es *experimentService) filterByLookupOrder(
 	filtered := matches
 	for _, segmenter := range segmenters {
 		orderedValues := filters[segmenter]
-		if len(orderedValues) == 1 {
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


treatment-service/services/experiment_service.go
```diff
@@ -128,38 +128,37 @@ func (es *experimentService) filterByLookupOrder(
 	filtered := matches
```
terryyylim, line 128: I'm wondering if the above `// Stop search when we have at least 1 match` is still relevant here. Feels like it should be shifted down 🤔

treatment-service/services/experiment_service.go
```diff
@@ -128,38 +128,37 @@ func (es *experimentService) filterByLookupOrder(
...
-			continue
-		}
-
 		currentFilteredList := []*models.ExperimentMatch{}
 		segmenterType := segmenterTypes[segmenter]
 		for _, transformedValue := range orderedValues {
 			if len(currentFilteredList) == 0 {
-				for _, segmenterMatch := range filtered {
-					segmenterMatchedValue := segmenterMatch.SegmenterMatches[segmenter]
+				for _, experiment := range filtered {
```
terryyylim, line 135: Thanks for renaming this 👍🏻 , `segmenterMatch` really didn't make any sense as I was reviewing these changes.

treatment-service/services/experiment_service.go
```diff
@@ -128,38 +128,37 @@ func (es *experimentService) filterByLookupOrder(
...
-							currentFilteredList = append(currentFilteredList, segmenterMatch)
+							currentFilteredList = append(currentFilteredList, experiment)
 						}
 					}
 				}
 			}
 		}
-		filtered = currentFilteredList
+		// currentFilteredList could be 0 in case of weak matches
+		if len(currentFilteredList) > 0 {
```
terryyylim, line 159: Just to affirm my understanding, this is because `orderedValues` could potentially be empty right?

terryyylim (approved):
Thanks for the changes, @krithika369! LGTM, left some comments/questions for clarification 🙏🏻

treatment-service/services/experiment_service.go
krithika369: I missed this comment - not sure it had much significance previously either. Removed it. Thanks for pointing out.

treatment-service/services/experiment_service.go
krithika369: Correct. Added more info in the comment now.

PR #33, opened 2022-09-02
Update UI dependencies

deadlycoconuts:
## Context
With the update of `mlp-ui` dependencies in https://github.com/gojek/mlp/pull/54 and https://github.com/gojek/mlp/pull/55 (in particular the update of `Node.js` **14 → 16**, `React` **16 → 17**, `@elastic/eui` **32.3.0 → 66.0.0** and `react-scripts` **to 5.0.1**), there is a need to similarly update XP UI, as it uses components from and is dependent on the `mlp-ui` package. 

## Main Modifications
- Replacement of the existing `Empty` component with the new `Page404` component
- Extensive adoption of the new `EuiPageTemplate` (see https://elastic.github.io/eui/#/templates/page-template) to manage page layouts; this replaces the existing usage of `EuiPage`, `EuiPageHeader`, `EuiPageSection`, etc. components
- Update of the existing colour palette manually specified to align with those of `v66` of `@elastic/eui` (see https://eui.elastic.co/#/theming/colors/values)
- Removal of redundant CSS styling used to accommodate `EuiFlyout`s
- Replacement of the caret toggle used inn Segmenter cards with a gear
Screenshot 2022-09-05 at 6 39 43 PM

#### Before
**Experiment/Treatment/Segment Details View:**&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; **Experiment/Treatment/Segment Edit View:**
.  
#### After
**Experiment/Treatment/Segment Details View:**&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; **Experiment/Treatment/Segment Edit View:**
.  

## Known Issues (Unable to be addressed in this PR)
- To follow up on `mlp-ui` to fix the `StepsWizardHorizontal` component as it does not seem to be highlighting as users progress through the timeline steps:
Screenshot 2022-09-05 at 9 37 21 AM

PR #36, opened 2022-09-16
Add additional experiment metadata to Fetch Treatment response

krithika369:
This PR adds additional metadata information to the Fetch Treatment response.

```javascript
{
    "data": {
        "experiment_id": 2,
        "experiment_name": "expertiment-2-sw",
        "metadata": { // Newly added field
            "experiment_type": "Switchback",
            "experiment_version": 2,
            "switchback_window_id": 1 // Only set for switchback experiments, starts from 0.
        },
        "treatment": {
            "configuration": {
                "test": "val"
            },
            "name": "treatment-b",
            "traffic": 0
        }
    }
}
```

The version field is now added to the database model for experiments, to reliably store the version of the experiment. Consequently, this field is also added to the Experiment Details view.

## Summary of Changes
The main changes are summarised below.

### Schema
* `api/proto/experiment.proto` - Add experiment version info in the Pubsub message
* `api/schema.yaml` - Add experiment version info in the `Experiment` structure, `SelectedTreatmentMetadata` into the treatment response structure

### Management Service
* `management-service/database/db-migrations/000003_experiment_version.*.sql` - Migrations for the `version` column in the `experiments table`. The `experiment_history` table already has this field.
* `management-service/models/experiment.go` - Include the version info when converting the DB model to other formats
* `management-service/services/experiment_history_service.go` - When creating the history record, we can now just copy the version from the current experiment record.

### Treatment Service
* `treatment-service/controller/treatment.go` - Return additional experiment metadata in the final response when an experiment is matched
* `treatment-service/services/treatment_service.go` - Return the switchback window Id (where applicable) from the `GetTreatment` method
* `treatment-service/models/typeconverter.go` - Cater to the newly added version field in the experiment data

### UI
* `ui/src/components/version_badge/VersionBadge.js` - Add new component for use in experiment / experiment history details view
* `ui/src/experiments/details/ExperimentDetailsView.js`, `ui/src/experiments/history/details/ExperimentHistoryDetailsView.js` - Include version info as a badge

Screenshot 2022-09-16 at 11 27 25 AM

## Miscellaneous Bugfix
A recent PR (https://github.com/caraml-dev/xp/pull/33) refactored the segmenter card layout in the Edit Project Settings form. This caused a small regression - when a segmenter has multiple options for variable mappings, it should be expanded by default, in the Selected Segmenters column. If not, when a new card is dragged to the Selected Segmenters column, there is no visual feedback to the users that some info must be filled (the form submit will be blocked by the yup validation but we can't see where the error is either). This is addressed by adding back the `initialIsOpen` prop in `ui/src/settings/components/form/components/segmenter_section/SegmenterCard.js`.


Makefile
```diff
@@ -49,8 +49,8 @@ generate-api:
 	oapi-codegen -templates api/templates/chi -config api/management/server.conf api/experiments.yaml
 	oapi-codegen -config api/mockmanagement/server.conf api/experiments.yaml
 	oapi-codegen -config api/treatment/server.conf api/treatment.yaml
-	cd clients/management/ && mockery --name=ClientInterface --output=../../clients/testutils/mocks --filename=ManagementClientInterface.go
-	cd clients/treatment/ && mockery --name=TreatmentClientInterface --output=../../clients/testutils/mocks --filename=TreatmentClientInterface.go
+	cd clients/management/ && mockery --name=ClientInterface --output=../../clients/testutils/mocks/management --filename=ManagementClientInterface.go
+	cd clients/treatment/ && mockery --name=ClientInterface --output=../../clients/testutils/mocks/treatment --filename=TreatmentClientInterface.go
```
krithika369, line 53: Renaming the interface for consistency and moving both to nested packages.

management-service/database/db-migrations/000003_experiment_version.up.sql
```diff
@@ -0,0 +1,9 @@
+ALTER TABLE experiments ADD version integer NOT NULL DEFAULT 1;
```
krithika369, line 1: The default value automatically takes care of those experiments that have no history.

api/schema.yaml
```diff
@@ -21,6 +21,23 @@ components:
         configuration:
           type: object
           description: Custom configuration associated with the given treatment
+    SelectedTreatmentMetadata:
```
terryyylim, line 24: Since the parent schema is already prefixed with `SelectedTreatment`, should we keep this concise with `Metadata`?

management-service/database/db-migrations/000003_experiment_version.up.sql
```diff
@@ -0,0 +1,9 @@
...
+
+-- Set the version number for existing records
+WITH agg_data AS (
+    SELECT t1.id AS id, count(*) AS num_history
+    FROM experiments t1 INNER JOIN experiment_history t2 ON t1.id = t2.experiment_id
+    GROUP BY t1.id
+) UPDATE experiments SET version = num_history+1
```
terryyylim, line 8: nit: I think it's neater to start `UPDATE` on a newline.

management-service/controller/experiment_controller_test.go
```diff
@@ -81,7 +81,8 @@ func (s *ExperimentControllerTestSuite) SetupSuite() {
 			"type": "",
 			"start_time": "0001-01-01T00:00:00Z",
 			"updated_at": "0001-01-01T00:00:00Z",
-			"updated_by": ""
+			"updated_by": "",
+			"version": 0
```
terryyylim, line 85: nit: Since we expect versions to start from 1, shall we use 1 instead for these tests?

plugins/turing/runner/experiment_runner_test.go
```diff
@@ -180,6 +180,10 @@ func TestFetchTreatment(t *testing.T) {
 									},
 									"name": "test_experiment-control",
 									"traffic": 50
+								},
+								"metadata": {
+									"experiment_version": 1,
+									"experiment_type": "a/b"
```
terryyylim, line 186: I think the expected value should be `"A/B"` right? https://github.com/caraml-dev/xp/blob/main/api/schema.yaml#L368

ui/src/experiments/details/ExperimentDetailsView.js
```diff
@@ -105,18 +119,19 @@ const ExperimentDetailsView = ({ projectId, experimentId, ...props }) => {
                 </ExperimentActions>
               </EuiPageTemplate.Header>
             </Fragment>
-          )}
+          )
+          }
```
terryyylim, line 123: Formatting looks funny here 🤔

ui/src/experiments/details/ExperimentDetailsView.js
```diff
@@ -105,18 +119,19 @@ const ExperimentDetailsView = ({ projectId, experimentId, ...props }) => {
...
 
           <Router primary={false}>
             <Redirect from="/" to="details" noThrow />
             <ExperimentConfigView path="details" experiment={data.data} />
             <ListExperimentHistoryView path="history" experiment={data.data} />
             <EditExperimentView path="edit" experimentSpec={data.data} />
           </Router>
-        </Fragment>
+        </Fragment >
```
terryyylim, line 131: Weird spaces at the back here.

terryyylim (approved):
Overall LGTM, left some small comments. Thanks @krithika369 ! 🚀

api/schema.yaml
krithika369: That might not be advisable, since the schema definitions are shared. The field inside `SelectedTreatment` is just called `Metadata` though.

plugins/turing/runner/experiment_runner_test.go
krithika369: It's only a mocked test, but I updated it to match the treatment service response anyway. Thanks!

ui/src/experiments/details/ExperimentDetailsView.js
krithika369: Something was broken with my local hooks, fixed it now.

management-service/controller/experiment_controller_test.go
krithika369: I intentionally left that field (and previously other properties too) empty - they don't play a role in the test cases. So they are empty values by default.

krithika369:
Thanks for the review @terryyylim! Will merge this once the CI passes.

PR #37, opened 2022-09-19
Add standard ensembler UI

deadlycoconuts:
With the [changes](https://github.com/caraml-dev/turing/pull/235) introduced in Turing to support the usage of Standard Ensemblers with Custom Experiment Engines, the Turing Experiments UI has been updated in this PR to expose Standard Ensembler components consumed by the Turing UI. 

There are two new components that Turing Experiments UI will now expose - `EditStandardEnsemblerConfig` and `StandardEnsemblerConfigDetails`. As their names suggest, they are used respectively for the create/edit router form and the router details view in the Turing UI respectively. 

**StandardEnsemblerConfigDetails**
Screenshot 2022-09-21 at 1 53 03 PM

**EditStandardEnsemblerConfig**
Screenshot 2022-09-21 at 5 44 32 PM

**Sample User Workflow**

https://user-images.githubusercontent.com/36802364/191474653-ee275e3c-cc6e-40d8-a3c0-ef7d695eb55f.mov

In addition, the `appConfig` now has a new `routeNamePathPrefix` field that specifies the path prefix that gets appended to a user-defined treatment configuration. This path prefix reflects the nesting of the treatment configuration within the final response payload that the treatment service, and/or subsequently the Turing Experiments plugin, sends back to the client.

For example, if the user-defined treatment configuration is:
```json
{
    "route_name": "control"
    ...
}
```
and the client response is as follows:
```json
{
    "treatment": {
        "configuration": {
            "route_name": "control"
            ...
        }
        ....
    }
    ...
}
```
then the path prefix will have to be specified as `treatment.configuration.` (including the final period).

## Main Modifications
- `ui/craco.config.js` - Addition of the exposed component paths to be consumed by the `ModuleFederationPlugin`
- `ui/src/config.js` - Addition of the `routeNamePathPrefix` field to the `appConfig`
- `ui/src/turing/components/configuration/StandardEnsemblerConfigDetails.js` - Addition of the new `StandardEnsemblerConfigDetails` component to be exposed
- `ui/src/turing/components/form/EditStandardEnsemblerConfig.js` - Addition of the new `EditStandardEnsemblerConfig`


ui/src/turing/components/configuration/StandardEnsemblerConfigDetails.js
```diff
@@ -0,0 +1,45 @@
+import React from "react";
+import { ConfigProvider, useConfig } from "config";
+import { EuiFlexGroup, EuiFlexItem } from "@elastic/eui";
+import { ConfigSectionPanel } from "components/config_section/ConfigSectionPanel";
+import { LinkedRoutesTable } from "../form/standard_ensembler/LinkedRoutesTable";
```
terryyylim, line 5: Let's use absolute path for all imports.

ui/src/turing/components/configuration/standard_ensembler_config/RouteNamePathConfigGroup.js
```diff
@@ -0,0 +1,29 @@
...
+} from "@elastic/eui";
+
+export const RouteNamePathConfigGroup = ({ routeNamePath }) => {
+  return (
+    <EuiTitle size="xs">
+      <EuiTextColor>
+        <EuiDescriptionList
+          textStyle="reverse"
+          type="responsiveColumn"
+          compressed>
```
terryyylim, line 18: Shall we use a map format for this instead? 
```
const routeNamePathConfig = [
  {
    title: 'Route Name Path',
    description: routeNamePath,
  }
]
```

With the `textStyle="reverse"` setting, I think the title shouldn't have been bold? Hence, was wondering if the above format would help achieve this instead.
image

ui/src/turing/components/configuration/standard_ensembler_config/RouteNamePathConfigGroup.js
```diff
@@ -0,0 +1,29 @@
...
+  EuiDescriptionListDescription,
+  EuiDescriptionListTitle,
+  EuiTextColor,
+  EuiTitle,
+} from "@elastic/eui";
+
+export const RouteNamePathConfigGroup = ({ routeNamePath }) => {
+  return (
+    <EuiTitle size="xs">
+      <EuiTextColor>
```
terryyylim, line 14: What happens if `EuiTextColor` is removed here?

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
```diff
@@ -0,0 +1,86 @@
...
+
+export const LinkedExperimentsContextMenu = ({
+  item,
+  projectId,
+  routeToExperimentMappings,
+  isButtonPopoverOpen,
+  setIsButtonPopoverOpen,
+  experimentStatus
+}) => {
+  let numRoutes = routeToExperimentMappings[item.id] ? Object.keys(routeToExperimentMappings[item.id][experimentStatus]).length : 0;
```
terryyylim, line 19: Should this be `numExperiments`? 🤔

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
```diff
@@ -0,0 +1,86 @@
...
+        initialPanelId={0}
+        panels={
+          [
+            {
+              id: 0,
+              title: `Linked ${experimentStatus[0].toUpperCase() + experimentStatus.slice(1)} Experiments`,
+              items: Object.values(routeToExperimentMappings[item.id][experimentStatus]).map(e => (
+                {
+                  name: (
+                    <EuiTextColor>
```
terryyylim, line 64: Is `EuiTextColor` necessary here? Rather than using `html a` tag, can we use `EuiLink` instead?

ui/src/turing/components/form/standard_ensembler/RouteNamePathFlyout.js
```diff
@@ -0,0 +1,81 @@
...
+    "route_name": "control",
+    ...
+}`
+                }
+              </EuiCodeBlock>
+
+              <p>
+                the client response that gets sent back to the Turing Router is actually as follows:
+              </p>
+              <EuiCodeBlock language="js" fontSize="m" paddingSize="m">
```
terryyylim, line 56: Should the language be "json"?

terryyylim (changes requested):
LGTM on the whole, left some comments to be addressed, thank you @deadlycoconuts !

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
```diff
@@ -0,0 +1,166 @@
...
+          projectId={projectId}
+          routeToExperimentMappings={routeToExperimentMappings}
+          isButtonPopoverOpen={isButtonPopoverOpen}
+          setIsButtonPopoverOpen={setIsButtonPopoverOpen}
+          experimentStatus={"running"}
+        />
+      ),
+    },
+    {
+      field: "scheduled_experiments",
```
terryyylim, line 136: How is this "scheduled_experiments" key computed and used by the `EuiInMemoryTable`?

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
```diff
@@ -0,0 +1,86 @@
...
+      iconSide={"right"}
+      color={"primary"}
+      onClick={onButtonClick}
+      isDisabled={numRoutes === 0}
+    >
+      {numRoutes}
+    </EuiButtonEmpty>
+  );
+
+  return isButtonPopoverOpen[item.id] ? (
```
terryyylim, line 46: Let's use `&&` and not have a ternary that returns `null` here.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
```diff
@@ -0,0 +1,166 @@
...
+      }
+    }
+  }, [isLoaded, experiments, paging, isAllExperimentsLoaded]);
+
+  const columns = [
+    {
+      field: "status",
+      width: "5px",
+      render: (_, item) => {
+        const isAssigned = routeToExperimentMappings[item.id] ?
```
terryyylim, line 96: `status` and `route_name` column computations seem to be duplicated here. Would it make sense to pre-aggregate all column values prior to passing them into the `items` prop of `EuiInMemoryTable`?

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
deadlycoconuts: Thanks for catching that! I've just edited it!

ui/src/turing/components/form/standard_ensembler/RouteNamePathFlyout.js
deadlycoconuts: Oops yep, I probably subconsciously read `js` as an acronym for "jay-sawn" and didn't notice anything wrong with it 😅

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
deadlycoconuts: Wow that's an awesome catch! Yes it's meant to be `numExperiments`; thanks!

ui/src/turing/components/configuration/standard_ensembler_config/RouteNamePathConfigGroup.js
deadlycoconuts: Actually, nothing. I copied this component off somewhere and didn't remove it 😅

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
deadlycoconuts: `EuiTextColor`: Oops not necessary, forgot to remove that. As for `EuiLink`, why would you prefer it to the `a` tag? I can use either one, no real preference except for some additional free styling we get.

ui/src/turing/components/configuration/standard_ensembler_config/RouteNamePathConfigGroup.js
deadlycoconuts: Hmm I think I was going for the same look as this total number of experiments look but I guess we can make it look like the fallback panel too:

Screenshot 2022-09-22 at 6 46 03 PM

deadlycoconuts: And no apparently it seems like changing to the map format does not work either. It's probably got to do with some lower-level styling because the panel showing the total number of experiments is like that too:

Screenshot 2022-09-22 at 6 52 03 PM

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: All the columns (with the exception of `status`) actually use multiple column values (of the rows in `items`, which represents the routes configured) to render the cells of the table. While it's easy to precompute the cells for `status` (`isAssigned` is the only value needed to determine `status`) and store them as a column, `route_name` for example needs both the value of `isAssigned` as well as the `id` of each route. 

This poses a problem because it's not possible to pass 2 (or more) values into the `field` prop of `EuiTableFieldDataColumnType` for the `EuiBasicTable` to extract both values and then pass them as arguments to the function specified in the `render` prop. This means that whatever precomputed values aggregated in advance would only be useful for the partially rendering each cell, because the function of the `render` prop would still have to access or compute other values from each row on-the-fly to determine how the cell ultimately gets rendered.

That said, I think this makes aggregating the computations in advance less useful than expected, since we aren't just extracting a single value to display for each cell, and still relying on the `render` prop to perform the rendering. 

On the other hand, I've not tried precomputing each column as a `ReactNode` component and then getting the table to extract these values as is (without having to rely on the `render` function) i.e. having the `field`-s set as `"column_1"`, `"column_2"`, `"column_3"` ..., and then having these keys within each of the elements in `items` corresponding to `<Column1... />`,  `<Column2... />`,  `<Column3... />`, ... 🤔

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: They aren't 😅 It's just a dummy and ultimately all the values get computed on the fly by the function passed to the `render` prop.

ui/src/turing/components/configuration/StandardEnsemblerConfigDetails.js
deadlycoconuts: Should I keep the relative paths just for imports within the same folder (`./**`)? It seems like we are sometimes keeping the relative paths (to supposedly better reflect the hierarchy of the imported components) and other times preferring absolute paths 👀

krithika369: I remember we discussed this a while ago on a different PR. I don't think there is a strong convention around it. In general, I think the absolute imports are convenient to work with especially for importing shared components (you don't have to think about where a component is located in relation to the current component). So I'd definitely avoid `../`.

But when we are importing a component that's nested within the current component's structure, I think relative imports are ok - in fact, you can make the reverse argument now (easier to do `./EnsemblerDetailsRow` than `turing/components/configuration/StandardEnsembler/EnsemblerDetailsRow` or something like that). The rule of thumb I'm trying to use recently is - do the 2 components have a composition kind of relationship, where you can always say that one will only be used by the other (=> if the parent moves, the child component will just move with it)? If so, I think `./` is ok (and may be preferable). `../` is a strict no. 😅

ui/src/services/experiment/ExperimentStatus.js
```diff
@@ -13,7 +13,7 @@ export const getExperimentStatus = (experiment) => {
   const statusMapping = {
     Deactivated: {
       label: "Deactivated",
-      color: "subdued",
+      color: "#6A717D",
```
deadlycoconuts, line 16: Fixed this colour which is no longer supported as part of the EUI colour palette; it now needs to be specified manually with a colour hex.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
```diff
@@ -0,0 +1,166 @@
...
+import { LinkedExperimentsContextMenu } from "./LinkedExperimentsContextMenu";
+
+export const LinkedRoutesTable = ({
+  projectId,
+  routes,
+  treatmentConfigRouteNamePath,
+}) => {
+  const { appConfig } = useConfig();
+
+  const [isButtonPopoverOpen, setIsButtonPopoverOpen] = useState(routes.reduce((m, r) => {m[r.id] = {running: false, scheduled: false}; return m}, {}));
```
krithika369, line 25: Hmmm having this state in the parent component makes the implementation a little more complex and less coupled with the button that it belongs to. Can this be defined directly inside `LinkedExperimentsContextMenu` component so that each clickable item has its own state? And can it simply track `true/false` ? Then, if any of them are true, we can use the `routeToExperimentMappings` to list the experiments in the popover, for that specific status - right?

We will be creating more state variables but since they are all lean, I don't think that should be a problem.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
```diff
@@ -0,0 +1,166 @@
...
+    },
+    {
+      field: "scheduled_experiments",
+      width: "35%",
+      name: "Scheduled Experiments",
+      render: (_, item) => (
+        <LinkedExperimentsContextMenu
+          item={item}
+          projectId={projectId}
+          routeToExperimentMappings={routeToExperimentMappings}
```
krithika369, line 143: Is it enough to send `routeToExperimentMappings[item.id][status]` into this component as `linkedExperiments` ?

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
```diff
@@ -0,0 +1,166 @@
...
+    [appConfig]
+  );
+
+  const [{ data: { data: experiments, paging }, isLoaded }] = useXpApi(
+    `/projects/${projectId}/experiments`,
+    {
+      query: {
+        start_time: start_time,
+        end_time: end_time,
+        page: pageIndex + 1,
```
krithika369, line 47: Shall we abstract this into a context function under `ui/src/providers/experiments`, say - `getAllExperiments(status)`? It will help keep this file more streamlined.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: I remember making a conscious choice to manage all the states of the parent because creating the individual states within `LinkedExperimentsContextMenu` somehow threw some error or minimally a warning regarding the creation of individual states by looping through with the function in `render`. I'll try to reproduce the issue again and share that error again.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: True that; that should be sufficient 😅

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: While we're on this, should we standardise the naming of the context providers we have? Should they be singular or plural? There's currently a mix of both and I'm not if it was intentional to have them that way e.g. we have a `SegmenterContextProvider` in a `segmenters` folder, but a `TreatmentsContextProvider` in a `treatment` folder 👀

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
deadlycoconuts: Correction: I remembered I need `null` as a return value because a function passed to the `render` prop needs to return minimally a `null` response as opposed to `undefined` which is what may happen if you use `&&` 
Screenshot 2022-09-23 at 4 20 55 PM

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
krithika369: Haha.. would be nice to standardarise. Since singular is more commonly used, let's use that.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: Okay I can't seem to get the error to reappear, maybe it was something else :/ I've made the changes to track the state of each button within each `LinkedExperimentsContextMenu`, and we'll be using `useToggle` to get the `true/false` state of the button.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: Just a slight update on this, I think we can just use `id` in the `field` for all of the columns, and then get the `render` function to do something with the `id` since we're almost always using the `routeToExperimentMappings` with the route name.

deadlycoconuts:
Thanks everyone for all the comments! I should've most of them covered. There's some significant refactoring done to `LinkedRouteTable` and `LinkedExperimentsContextMenu`, so it'd be nice to take a look at that!

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
terryyylim: I think considering that we're using `eui` libraries and React, it's cleaner to use these components rather than html tags unless we need to do something specific/not achievable via these components.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
```diff
@@ -0,0 +1,116 @@
...
+          experimentStatus={"scheduled"}
+        />
+      )
+    },
+  ];
+
+  return isAllExperimentsLoaded ? (
+    <EuiFlexItem>
+      <EuiInMemoryTable
+        items={routes.filter((r) => r.id !== "")}
```
terryyylim, line 105: In what cases, would `r.id` be an empty string?

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
```diff
@@ -0,0 +1,68 @@
...
+      iconSide={"right"}
+      color={"primary"}
+      onClick={togglePopover}
+      isDisabled={numExperiments === 0}
+    >
+      {numExperiments}
+    </EuiButtonEmpty>
+  );
+
+  return linkedExperiments ? (
```
terryyylim, line 34: In relation to a previous thread - https://github.com/caraml-dev/xp/pull/37#discussion_r978387554, since the rendering is dependent on `linkedExperiments` which is passed down from the parent component, why not do the following instead? Then, we won't need the ternary for this return statement.
```
render: (id) => routeToExperimentMappings[id] && (
        <LinkedExperimentsContextMenu
          projectId={projectId}
          linkedExperiments={routeToExperimentMappings[id]}
          experimentStatus={"scheduled"}
        />
      )
```

ui/src/turing/components/form/standard_ensembler/RouteNamePathFlyout.js
```diff
@@ -0,0 +1,81 @@
...
+        }
+        ....
+    }
+    ...
+}`
+                }
+              </EuiCodeBlock>
+
+              <p>
+                Hence, the path prefix is automatically specified as `treatment.configuration.` (including the final period).
```
terryyylim, line 71: Suddenly thought of this while reviewing again. I think `treatment.configuration` should not be hardcoded, but rather be a templated value retrieved from the specified configs.

ui/src/turing/components/form/standard_ensembler/RouteNamePathFlyout.js
```diff
@@ -0,0 +1,81 @@
...
+                treatment configuration.
+              </p>
+
+              <p>
+                This path prefix reflects the nesting of the treatment configuration within
+                the final response payload that the Turing Router finally receives.
+              </p>
+
+              <p>
+                In Turing Experiments' case, if the user-defined treatment configuration is:
```
terryyylim, line 43: Should `Turing Experiments` be `XP`? There are other parts in this PR that uses `Turing Experiments` too, wdyt? @krithika369

terryyylim (approved):
Thanks @deadlycoconuts , PR looks cleaner now! I left some small comments - feel free to merge after addressing them 😄

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: When a user clicks on `+ Add Route` and then doesn't type anything in the route name 👀 Or when he/she erases everything he/she has typed for a route name.

ui/src/turing/components/form/standard_ensembler/RouteNamePathFlyout.js
deadlycoconuts: Wow oh yes, nice job catching this!

ui/src/turing/components/form/standard_ensembler/RouteNamePathFlyout.js
deadlycoconuts: 👀👀👀

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
terryyylim: Should be addressed by this suggestion - https://github.com/caraml-dev/xp/pull/37#discussion_r979531215.

ui/src/turing/components/form/standard_ensembler/LinkedExperimentsContextMenu.js
deadlycoconuts: Thanks for the suggestion! Just made the change!

ui/src/turing/components/form/standard_ensembler/RouteNamePathFlyout.js
krithika369: Nope.. the `ui/src/turing/components` modules are meant for Turing. And XP should advertise itself as Turing Experiments. So it's intentional.

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
krithika369: Wait.. why was this resolved? Are we not moving this logic to get all experiments into a context function?

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
```diff
@@ -0,0 +1,166 @@
...
+      },
+    },
+    { data: [], paging: { total: 0 } }
+  );
+
+  const getRouteName = (config, path) => path.split('.').reduce((obj, key) => obj && obj[key], config);
+
+  // this stringified value of routes below allows the React effect below to mimic a deep comparison when changes to the
+  // array routes are made
+  const stringifiedRoutes = JSON.stringify(routes)
```
krithika369, line 29: Would it be better to use the ids only?

```javascript
const stringifiedRoutes = routes.map(e => e.id).join()
```

ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js
deadlycoconuts: I think you're viewing an outdated commit :o

krithika369: :O :O my bad.

deadlycoconuts:
It seems like all of the issues on this PR has been resolved; I'll be merging this in a short while! Thanks a lot everyone for your time! :)

PR #39, opened 2022-09-27
Addition of text wrapping style to route selection panel 

deadlycoconuts:
This PR introduces a tiny fix to the UI to allow the entire route name path of a standard ensembler to be displayed, instead of being truncated.

It involves adding an additional css style property to the existing `experimentSummaryPanel` class to allow text wrapping.

**Before:**
Screenshot 2022-09-27 at 4 27 32 PM

**After:**
Screenshot 2022-09-27 at 4 27 13 PM

### DIFF
--- a/ui/src/turing/components/configuration/experiments_config/ExperimentsConfigGroup.scss
+++ b/ui/src/turing/components/configuration/experiments_config/ExperimentsConfigGroup.scss
@@ -6,6 +6,7 @@
 
     .euiDescriptionList__description {
       width: calc(100% - 150px);
+      white-space: normal !important;
     }
   }
 }


ui/src/turing/components/configuration/StandardEnsemblerConfigDetails.js
```diff
@@ -14,7 +14,7 @@ const StandardEnsemblerConfigDetailsComponent = ({ projectId, routes, routeNameP
     <ConfigProvider>
       <EuiFlexGroup direction="row" wrap>
         <EuiFlexItem grow={1} className="euiFlexItem--smallPanel">
-          <ConfigSectionPanel title="Route Selection" className="experimentSummaryPanel">
```
krithika369, line 17: Do you think we can simply apply the `white-space: normal !important;` to the existing `experimentSummaryPanel` style ? Do you see any downside?

deadlycoconuts: Let me hunt around and see where the existing style is used 👀

deadlycoconuts: Seems like it's only used here: 
Screenshot 2022-09-27 at 5 05 11 PM

And since the number of experiments won't be anything as long as the route name path, I guess we can safely apply that to the `experimentSummaryPanel` as you suggested 🙂

krithika369 (approved):
LGTM!

PR #43, opened 2022-10-12
Make user-friendly status filter multi-search

krithika369:
**What this PR does / why we need it**: This is a follow up from the previous PR https://github.com/caraml-dev/xp/pull/42 which introduced a search filter for the experiments using user-friendly status names. However, in that PR, the API was only designed to take in a single value. This PR makes it possible to search using multiple values of the status, to match any of them. This required the underlying `gorm` library to be upgraded, as the OR clauses were not gracefully handled in the older version.

### Illustration
https://user-images.githubusercontent.com/23465343/195236447-d585e419-5220-499a-83b1-8c4a1c2055a3.mov

### Search Filter Changes
* `api/experiments.yaml` - Accept multiple values for `status_friendly`, in List Experiments filters
* `management-service/controller/experiment_controller.go` - Handle multiple values for `status_friendly` in List API search filters
* `management-service/services/experiment_service.go` - Add `filterExperimentStatusFriendly` to handle matching the status values.
* `ui/src/experiments/list/search/components/ExperimentStatusFilter.js` - Make the status filter multiselect

### GORM upgrade
* Move from:
    - `github.com/jinzhu/gorm` -> `gorm.io/gorm` @ latest
    - `github.com/jinzhu/gorm/dialects/postgres` -> `gorm.io/driver/postgres` @ latest
* `management-service/database/database.go`, `management-service/server/server.go`, `management-service/internal/testutils/database.go` - DB init / close related API changes
* `management-service/models/experiment.go`, `management-service/models/experiment_history.go` - Add `AfterFind` hook as the new ORM version doesn't respect the timezone info in the connection string.
* Some `int` / `int32` / `int64` conversions in OFFSET / LIMIT / COUNT queries to match the new APIs
* The `save` function in every service that saves the data to the DB is refactored to use Upsert instead of Create / Update. This is needed because the `NewRecord` function from Gorm v1 no longer exists. Ref: https://gorm.io/docs/create.html#Upsert-x2F-On-Conflict


management-service/services/experiment_service.go
```diff
@@ -153,10 +141,10 @@ func (svc *experimentService) ListExperiments(
 			// * the end_time parameter should fall within the experiment's (start and end) times
 			// * the experiment starts and ends within the [start_time and end_time) duration
 			query = query.Where(
-				`tstzrange(start_time, end_time, '[)') @> tstzrange(?, ?, '[]')
-				OR tstzrange(start_time, end_time, '()') @> tstzrange(?, ?, '[]')
-				OR tstzrange(?, ?, '[]') @> tstzrange(start_time, end_time, '[)')`,
-				params.StartTime, params.StartTime, params.EndTime, params.EndTime, params.StartTime, params.EndTime,
+				svc.query().
```
krithika369, line 144: Refactoring this OR clause to use the APIs rather than raw query string, since the new version of the Gorm API supports it correctly.

management-service/services/experiment_service.go
```diff
@@ -118,22 +119,8 @@ func (svc *experimentService) ListExperiments(
...
-				query = query.Where("start_time > ?", currentTime)
-			case ExperimentStatusFriendlyCompleted:
-				query = query.Where("end_time < ?", currentTime)
-			default:
-				// Status Running - current time should be present in the experiment duration.
-				query = query.Where("tstzrange(start_time, end_time, '[)') @> tstzrange(?, ?, '[]')", currentTime, currentTime)
-			}
-		}
+	if len(params.StatusFriendly) > 0 {
+		query = svc.filterExperimentStatusFriendly(query, params.StatusFriendly)
```
deadlycoconuts, line 123: Thanks for refactoring this into a separate method!

deadlycoconuts (approved):
LGTM! 🚀 Thanks a lot for updating all the instances where we're using the gorm library!

PR #44, opened 2022-10-17
Add field selector to list experiments API

deadlycoconuts:
This PR introduces a field selector to the `ListExperiments` API of the management service; previously, field selectors only existed for the `ListTreatments` and `ListSegments` APIs. 

The introduction of this field selector would allow more selective control over the fields of the experiment objects to return. Doing so would coincidentally allow the `ListExperimentsTable` of the UI to display larger numbers of experiments (>10) in each page, which it was previously unable to do because querying the API for an entire list of experiments would have resulted in too much data being transferred (due to the unnecessary fields/columns being returned). 

As such, this PR also makes a tiny change to the UI to activate the option to allow users to select the number of experiments they wish to display in each page of the table. This change also affects the `ListTreatmentsTable` and `ListSegmentsTable`.

ezgif com-gif-maker (1)

### Modifications
- `api/experiments.yaml` - Addition of the `fields` query field to the `ListExperiments` API
- `api/schema.yaml` - Addition of the `ExperimentField` schema 
- `management-service/models/experiment.go` - Modification of the `ToApiSchema` method to return only selected fields
- `management-service/pagination/pagination.go` - Modification of the existing `MaxPageSize` from 10 to 50
- `management-service/services/experiment_service.go` - Modification of the `ListExperiments` method to handle the fields, and minor refactoring of the start/end time filter handling into a separate helper method
- `ui/src/components/table/BasicTable.js` - Toggling of the prop to show page options in experiment/treatment/segment tables
- `ui/src/config.js` - Addition of new configuration to specify the number of pages/fields to query when using the `ListExperiments` API
- `ui/src/turing/components/form/standard_ensembler/LinkedRoutesTable.js` - Minor refactoring of the existing process to retrieve the status (friendly) of an experiment

### DIFF
--- a/management-service/services/experiment_service.go
+++ b/management-service/services/experiment_service.go
@@ -115,17 +115,11 @@ func (svc *experimentService) ListExperiments(
 	var exps []*models.Experiment
 
 	query := svc.query()
-	if params.Fields != nil && len(*params.Fields) != 0 {
-		err = validateListExperimentFieldNames(*params.Fields)
-		if err != nil {
-			return nil, nil, err
-		}
-		// query.Select only accepts []string
-		var fieldNames []string
-		for _, field := range *params.Fields {
-			fieldNames = append(fieldNames, string(field))
-		}
-		query = query.Select(fieldNames)
+
+	// Handle Field values
+	query, err = svc.filterFieldValues(query, params)
+	if err != nil {
+		return nil, nil, err
 	}
 
 	query = query.
@@ -136,11 +130,17 @@ func (svc *experimentService) ListExperiments(
 	if params.Status != nil {
 		query = query.Where("status = ?", params.Status)
 	}
-	if len(params.StatusFriendly) > 0 {
-		query = svc.filterExperimentStatusFriendly(query, params.StatusFriendly)
+
+	// Handle StatusFriday values
+	query = svc.filterStatusFriendlyValues(query, params)
+	if err != nil {
+		return nil, nil, err
 	}
-	if params.StartTime != nil && !params.StartTime.IsZero() && (params.EndTime == nil || params.EndTime.IsZero()) {
-		return nil, nil, errors.Newf(errors.BadInput, "end_time parameter must be supplied as well")
+
+	// Handle Start and EndTime values
+	query, err = svc.filterStartEndTimeValues(query, params)
+	if err != nil {
+		return nil, nil, err
 	}
 	if params.EndTime != nil && !params.EndTime.IsZero() && (params.StartTime == nil || params.StartTime.IsZero()) {
 		return nil, nil, errors.Newf(errors.BadInput, "start_time parameter must be supplied as well")
--- a/management-service/services/experiment_service.go
+++ b/management-service/services/experiment_service.go
@@ -166,7 +166,7 @@ func (svc *experimentService) ListExperiments(
 	// Pagination
 	var pagingResponse *pagination.Paging
 	var count int64
-	if params.Fields == nil || params.Page != nil || params.PageSize != nil {
+	if params.Fields == nil {
 		err = pagination.ValidatePaginationParams(params.Page, params.PageSize)
 		if err != nil {
 			return nil, nil, err
--- a/management-service/services/experiment_service.go
+++ b/management-service/services/experiment_service.go
@@ -558,8 +558,8 @@ func (svc *experimentService) filterFieldValues(query *gorm.DB, params ListExper
 			return nil, err
 		}
 
-		// Stores a map of unique field names which are unique column names to be selected in the db query
-		fieldNamesSet := make(map[string]struct{})
+		// Stores a set of unique field names which are unique column names to be selected in the db query
+		fieldNamesSet := set.New()
 		for _, field := range *params.Fields {
 			fieldName := field
 			// Add ExperimentFieldStatus to the query because status_friendly does not exist in the db as a column
@@ -567,19 +567,21 @@ func (svc *experimentService) filterFieldValues(query *gorm.DB, params ListExper
 				// Add ExperimentFieldStartTime and ExperimentFieldEndTime to the query as they are required for
 				// determining the statusFriendly field; these two fields may or may not be returned depending on
 				// what is actually specified in params.Fields
-				fieldNamesSet[string(models.ExperimentFieldStartTime)] = struct{}{}
-				fieldNamesSet[string(models.ExperimentFieldEndTime)] = struct{}{}
+				fieldNamesSet.Insert(string(models.ExperimentFieldStartTime))
+				fieldNamesSet.Insert(string(models.ExperimentFieldEndTime))
 
 				fieldName = models.ExperimentFieldStatus
 			}
-			fieldNamesSet[string(fieldName)] = struct{}{}
+			//fieldNamesSet[string(fieldName)] = struct{}{}
+			fieldNamesSet.Insert(string(fieldName))
 		}
 
 		// Retrieve a slice of unique strings from fieldNamesSet; query.Select only accepts []string
 		var fieldNames []string
-		for k := range fieldNamesSet {
-			fieldNames = append(fieldNames, k)
-		}
+		fieldNamesSet.Do(func(fieldName interface{}) {
+			fieldNames = append(fieldNames, fmt.Sprint(fieldName))
+		})
+
 		query = query.Select(fieldNames)
 	}
 	return query, nil
--- a/management-service/services/experiment_service.go
+++ b/management-service/services/experiment_service.go
@@ -166,22 +166,24 @@ func (svc *experimentService) ListExperiments(
 	// Pagination
 	var pagingResponse *pagination.Paging
 	var count int64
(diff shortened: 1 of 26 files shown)


management-service/services/experiment_service.go
```diff
@@ -578,6 +588,35 @@ func (svc *experimentService) filterExperimentStatusFriendly(query *gorm.DB, sta
 	return query.Where(orPredicates)
 }
 
+func (svc *experimentService) filterStartEndTimeValues(query *gorm.DB, params ListExperimentsParams) (*gorm.DB, error) {
```
deadlycoconuts, line 591: Created a helper method out of this because the linter was throwing out an excessive cyclomatic complexity for the `ListExperiments` method.

management-service/services/experiment_service.go
```diff
@@ -172,7 +166,7 @@ func (svc *experimentService) ListExperiments(
 	// Pagination
 	var pagingResponse *pagination.Paging
 	var count int64
-	err := pagination.ValidatePaginationParams(params.Page, params.PageSize)
+	err = pagination.ValidatePaginationParams(params.Page, params.PageSize)
```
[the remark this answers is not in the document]
deadlycoconuts: Oh my good catch! Thanks for the reminder; can't remember why I left this out 😵‍💫

treatment-service/integration-test/fetch_treatment_it_test.go
```diff
@@ -811,5 +812,5 @@ func (suite *TreatmentServiceTestSuite) TestLocalStorage() {
 
 	resp, err := suite.managementServiceClient.GetExperimentWithResponse(suite.ctx, 1, 1)
 	suite.Require().NoError(err)
-	suite.Require().Equal(resp.JSON200.Data.Name, storage.Experiments[1][0].Experiment.Name)
+	suite.Require().Equal(storage.Experiments[1][0].Experiment.Name, *resp.JSON200.Data.Name)
```
deadlycoconuts, line 815: Reordered the 2 arguments because the expected argument is supposed to come before the actual one.

treatment-service/models/storage_test.go
```diff
@@ -110,7 +117,7 @@ func (suite *LocalStorageLookupSuite) SetupTest() {
 		Return(&http.Response{
 			StatusCode: 200,
 			Header:     map[string][]string{"Content-Type": {"json"}},
-			Body:       ioutil.NopCloser(bytes.NewBufferString(`{"data" : []}`)),
+			Body:       io.NopCloser(bytes.NewBufferString(`{"data" : []}`)),
```
deadlycoconuts, line 120: Replaced the deprecated method with its replacement.

management-service/services/experiment_service.go
```diff
@@ -555,6 +551,40 @@ func (svc *experimentService) save(exp *models.Experiment) (*models.Experiment,
...
+			}
+			fieldNamesSet[string(fieldName)] = struct{}{}
+		}
+
+		// Retrieve a slice of unique strings from fieldNamesSet; query.Select only accepts []string
+		var fieldNames []string
+		for k := range fieldNamesSet {
+			fieldNames = append(fieldNames, k)
+		}
+		query = query.Select(fieldNames)
```
deadlycoconuts, line 583: This `fieldFieldValues` function essentially specifies `SELECT` values in the DB query such that only columns corresponding to the field values requested get retrieved.

management-service/services/experiment_service.go
```diff
@@ -555,6 +551,40 @@ func (svc *experimentService) save(exp *models.Experiment) (*models.Experiment,
...
+			fieldName := field
+			// Add ExperimentFieldStatus to the query because status_friendly does not exist in the db as a column
+			if field == models.ExperimentFieldStatusFriendly {
+				// Add ExperimentFieldStartTime and ExperimentFieldEndTime to the query as they are required for
+				// determining the statusFriendly field; these two fields may or may not be returned depending on
+				// what is actually specified in params.Fields
+				fieldNamesSet[string(models.ExperimentFieldStartTime)] = struct{}{}
+				fieldNamesSet[string(models.ExperimentFieldEndTime)] = struct{}{}
+
+				fieldName = models.ExperimentFieldStatus
```
deadlycoconuts, line 573: This block is added since `status_friendly` does not exist in the DB as a column. In order to determine what this field value should be, we'll need to retrieve _other_ columns, such as `status`, `start_time` and `end_time`. 

The `fieldNamesSet` used to store column names ensure that only unique values are used in the query.

management-service/services/experiment_service_test.go
```diff
@@ -163,28 +163,30 @@ func testListExperiments(s *ExperimentServiceTestSuite) {
 	stringSegmenter := []string{"seg-1"}
 	boolSegmenter := []string{"true"}
 
+	var nilPagingResponse *pagination.Paging
+
 	// All experiments under a settings
-	expResponsesList, pagingResponse, err := svc.ListExperiments(1, services.ListExperimentsParams{})
+	actualResponsesList, pagingResponse, err := svc.ListExperiments(1, services.ListExperimentsParams{})
```
deadlycoconuts, line 169: Renamed this as it's clearly the actual response, and not the expected resopnse.

management-service/services/experiment_service.go
```diff
@@ -555,6 +551,40 @@ func (svc *experimentService) save(exp *models.Experiment) (*models.Experiment,
...
 
+func (svc *experimentService) filterFieldValues(query *gorm.DB, params ListExperimentsParams) (*gorm.DB, error) {
+	if params.Fields != nil && len(*params.Fields) != 0 {
+		err := validateListExperimentFieldNames(*params.Fields)
+		if err != nil {
+			return nil, err
+		}
+
+		// Stores a map of unique field names which are unique column names to be selected in the db query
+		fieldNamesSet := make(map[string]struct{})
```
krithika369, line 562: Why not use `"github.com/golang-collections/collections/set"` for this, which we commonly use elsewhere in the codebase?

ui/src/config.js
```diff
@@ -36,6 +36,11 @@ export const appConfig = {
     : [{ href: "https://github.com/caraml-dev/xp", label: "XP User Guide" }],
   pagination: {
     defaultPageSize: 10,
+    experimentContextPageSize: 50,
+  },
+  listExperimentFields: {
```
krithika369, line 41: I'm not sure this should be an app-level config. Can it just be embedded in the components that use them? Reasons:
* If this were an app-config, I think it would make sense to let it drive the behavior of the components entirely (like modifying the `experimentTableFields` here could also change the columns displayed in the experiments table, which is not the case now and I think that would be over-engineering at this point).
* These configs are not re-used in multiple places which could also be a hint that this can be localised.

If either of the above reasons changed, we could consider extracting them out again but perhaps we will need to think more about what influence such configs will have (bullet # 1 above).

ui/src/providers/experiment/context.js
```diff
@@ -23,38 +19,24 @@ export const ExperimentContextProvider = ({ projectId, children }) => {
...
-      }
-    }
-  }, [isLoaded, experiments, paging, isAllExperimentsLoaded]);
-
   return (
     <ExperimentContext.Provider
       value={{
-        allExperiments,
-        isAllExperimentsLoaded,
+        experiments,
```
krithika369, line 38: Should we rename this to `scheduledAndRunningExperiments` or something to be more specific? Although the `ExperimentContextProvider` is only used in the Turing microfrontend components now, a more accurate name could help development in the future if we look to use it in other places.

krithika369 (approved):
Left some small comments. LGTM. Thanks for this PR! Please feel free to merge omnce you've addressed the comments.

management-service/services/experiment_service.go
deadlycoconuts: Ahh, I tried doing that but I realised we can't extract the values contained within the `Set` as a slice of strings because the underlying field (`hash`) containing all the values added to the `Set` is actually a private field:

```go
type (
	Set struct {
		hash map[interface{}]nothing
	}

	nothing struct{}
)
```

ui/src/config.js
deadlycoconuts: That's true, I was thinking of #1 as a distant possibility but yeah now that you mention it, it's probably over-engineering; I've removed both configs and directly embedded the lists in the components they are used.

management-service/services/experiment_service.go
krithika369: Hmm there should be a function to get it out.

krithika369: https://github.com/golang-collections/collections/blob/604e922904d3/set/set.go#L36

deadlycoconuts: They seem to all return a `*Set` at most 😕 https://pkg.go.dev/github.com/golang-collections/collections/set

deadlycoconuts: An alternative is to run `Do()` with a function that stores all the string values in the set into some slice.

deadlycoconuts:
Thanks a lot for the comments @krithika369, I've updated the PR to address them and I'll be merging this right now!

PR #45, opened 2022-10-18
Migrate to react-router-dom

krithika369:
**What this PR does / why we need it**: This PR follows the changes implemented in https://github.com/gojek/mlp/pull/56, for the XP application.
1. Migration from `@reach/router` to `react-router-dom`, per the [official guidelines](https://reactrouter.com/en/v6.3.0/upgrading/reach).
2. `ui/src/AppRoutes.js` - Exposing all the routes under `AppRoutes`, as a federated component. Unlike with Turing, the routes are not exposed via module federation for embedding within the MLP shell app directly. Instead, we'll continue to use the `ExperimentsLandingPage` component, under Turing's Experiments view.


ui/src/services/segmenter/SegmenterStatus.js
```diff
@@ -2,7 +2,7 @@ export const getSegmenterStatus = (segmenter) => {
   const statusMapping = {
     inactive: {
       label: "Inactive",
-      color: "subdued",
+      color: "default",
```
krithika369, line 5: Subdued is no longer a valid badge color in the newer Eui versions.

ui/src/segments/details/SegmentDetailsView.js
```diff
@@ -62,49 +65,48 @@ const SegmentDetailsView = ({ projectId, segmentId, ...props }) => {
...
-                        actions={getActions(data.data)}
-                        selectedTab={props["*"]}
-                        {...props}
-                      />
-                    )}
-                  </SegmentActions>
-                </EuiPageTemplate.Header>
-              </Fragment>
-            )}
+      {!isLoaded ? (
```
deadlycoconuts, line 68: Minor nitpick: should we indent this block? Or rather in general do we prefer to indent such blocks when they are enclosed in tags?

deadlycoconuts (approved):
Looks good to me! 🚀 Thanks a lot for making all the necessary changes to migrate to React Router v6! Just left some tiny comments!

ui/src/experiments/ExperimentsLandingPage.js
```diff
@@ -21,62 +21,57 @@ import TreatmentDetailsView from "treatments/details/TreatmentDetailsView";
...
-          />
-          <CreateTreatmentView path="/treatments/create" />
-          <TreatmentDetailsView path="/treatments/:treatmentId/*" />
+          {/* TREATMENTS */}
+          <Route path="treatments">
+            <Route index element={<LandingView Component={ListTreatmentsView} name="Treatments" />} />
+            <Route path="create" element={<CreateTreatmentView />} />
+            <Route path=":treatmentId">
+              <Route path="history/:version" element={<TreatmentHistoryDetailsView />} />
+              <Route index path="*" element={<TreatmentDetailsView />} />
```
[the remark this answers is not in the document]
krithika369: I switched up the order of the index route. We need the `path="*"` because each component has a nested set of routes under it (eg: The treatment detail view has different tabs).

ui/src/segments/details/SegmentDetailsView.js
krithika369: I don't have a strong preference on this and from the search that I did, though ESlint and other formatters seem to support some configurations to enforce this, the general guideline is to keep it readable. So I suppose we could leave it as it is. Unless you have some other suggestions to bring about more consistency.

deadlycoconuts: Oh I don't really have a preference really, it's just that I thought I spotted another change made in this PR where a similar block was indented (for the reverse reason) 😅 I guess we can keep it as it is and then... fix some convention if it gets too inconsistent that we get annoyed by it 😶 Thanks a lot for looking up all the formatters though!

ui/src/experiments/ExperimentsLandingPage.js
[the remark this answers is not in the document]
deadlycoconuts: Ahh okay I see, alright then! :)

krithika369:
Thanks for the review, merging this PR!

PR #47, opened 2022-10-20
Refactor config passing to XP plugin

deadlycoconuts:
As part of an ongoing process to make the Treatment Service of XP runnable as a plugin for [Turing](https://github.com/caraml-dev/turing/) Router deployments, as opposed to calling a singular Treatment Service that has been deployed to receive treatments, this PR introduces a series of changes to pass configuration values required in the Treatment Service to the XP plugin. Subsequent PRs will feature other changes to actually serve treatments directly from the plugin and are thus outside the scope of this PR.

As there are quite a number of configuration values and secrets that the Treatment Service requires at start up, some of which are shared with the Management Service, while others are secrets which shouldn't be exposed directly, these values are made to be retrieved somewhat differently depending on which category they fall into:

a. Configuration values shared with the Management Service
```yaml
config:
    pub_sub:
        # ...
    segmenter_config:
        # ...
```
b. Configuration values unique to the Treatment Service
```yaml
config:
    new_relic_config:
        # ...
    sentry_config:
        # ...
    port:
        # ...
    management_service:
        # ...
    url:
        # ...
    deployment_config:
        # ...
    assigned_treatment_logger:
        # ...
    monitoring_config:
        # ...
    swagger_config:
        # ...
```
c. Secrets
- Google service account for PubSub communication with the Management Service and logging

Without going too deeply into the details of how the [Turing](https://github.com/caraml-dev/turing/) API operates, during each router deployment, the API will rely on the XP plugin manager to supply the configuration that needs to be deployed together with the XP plugin runner. In order to generate this configuration, the following steps need to be performed before returning the aforementioned experiment configuration to the Turing API (refer to the illustration below):

1. (XP plugin manager) Retrieves values shared with the Management Service via the HTTP GET endpoint at `\treatment-service-plugin-config`
2. (XP plugin manager) Use values stored within the XP plugin manager (these are passed to the plugin through Turing API's config file when the plugin manager gets initialised) and those retrieved in Step 1 to create the experiment configuration
3. (XP plugin manager) Return the created experiment configuration to the Turing API
4. (Turing API) Retrieve the secret(s) required 
5. (Turing API) Inject the required configuration values and secrets into the Turing Router deployment

xpplugin

As the responsibility of XP lies in performing **Steps 1, 2 and 3**, this PR specifically introduces changes to support those steps, while a separate [PR](https://github.com/caraml-dev/turing/pull/274) in Turing has been created to perform Steps 4 and 5.

#### Management Service
- `api/experiments.yaml` - Addition of new endpoint and request schemas to the OpenAPI specs
- `api/schema.yaml` - Addition of other new schemas to support the new endpoint to the OpenAPI specs
- `management-service/controller/configuration_controller.go` - Addition of a new configuration controller to handle the new endpoint
- `management-service/services/configuration_service.go` - Addition of a new configuration service to generate the treatment service plugin configs for the XP plugin manager

#### XP Plugin
- `plugins/turing/config/config.go` - Addition of new structs to support the passing of treatment service plugin configs via json between the plugin manager, plugin runner and the Turing API
- `plugins/turing/manager/experiment_manager.go` - Addition of new methods that are called in `GetExperimentRunnerConfig` to return configs that are required for the treatment service


common/testutils/utils.go
```diff
@@ -14,9 +15,53 @@ func ReadFile(filepath string) ([]byte, error) {
...
 	}
 	return byteValue, nil
 }
+
+// testSetupEnvForGoogleCredentials creates a temporary file containing dummy service account JSON
+// then set the environment variable GOOGLE_APPLICATION_CREDENTIALS to point to the file.
+// This is useful for tests that assume Google Cloud Client libraries can automatically find
+// the service account credentials in any environment.
+// At the end of the test, the returned function can be called to perform cleanup.
+func TestSetupEnvForGoogleCredentials(t *testing.T) (reset func()) {
```
deadlycoconuts, line 30: Relocated this from `management-service/services/mlp_service_test.go` because this really helpful test helper is now used elsewhere besides the `services` package 😅

management-service/services/segmenter_service_test.go
```diff
@@ -1176,7 +1175,7 @@ func NewCountrySegmenter(_ json.RawMessage) (segmenters.Segmenter, error) {
 	return segmenters.NewBaseSegmenter(countryConfig), nil
 }
 
-func NewAreaSegmenter(configData json.RawMessage) (segmenters.Segmenter, error) {
+func NewAreaSegmenter(_ json.RawMessage) (segmenters.Segmenter, error) {
```
deadlycoconuts, line 1178: Minor refactoring of the method argument because it isn't even used at all in the function (maybe we can even simplify the signature?)

plugins/turing/config/config.go
```diff
@@ -31,20 +31,117 @@ type RemoteUI struct {
...
-	Enabled        bool           `json:"enabled"`
-	BaseURL        string         `json:"base_url"`      // Base URL for XP experiment REST API
-	HomePageURL    string         `json:"home_page_url"` // Website URL for end-users to manage experiments
-	RemoteUI       RemoteUI       `json:"remote_ui"`
-	RunnerDefaults RunnerDefaults `json:"runner_defaults"`
+	Enabled                      bool                         `json:"enabled"`
+	BaseURL                      string                       `json:"base_url"`      // Base URL for XP experiment REST API
+	HomePageURL                  string                       `json:"home_page_url"` // Website URL for end-users to manage experiments
+	RemoteUI                     RemoteUI                     `json:"remote_ui"`
+	RunnerDefaults               RunnerDefaults               `json:"runner_defaults"`
```
krithika369, line 38: Perhaps this change would happen after the plugin's experiment runner has been refactored too. But confirming:
Do we need the `RunnerDefaults` here anymore? Since the endpoint under this struct will no longer be needed with the plugin implementation anymore, shall we remove this struct and simply place the Timeout value under `TreatmentServicePluginConfig `?

plugins/turing/config/config.go
```diff
@@ -31,20 +31,117 @@ type RemoteUI struct {
...
+	SentryConfig            SentryConfig                  `json:"sentry_config"`
+	DeploymentConfig        DeploymentConfig              `json:"deployment_config"`
+	PubSub                  PubSub                        `json:"pub_sub"`
+	ManagementService       ManagementServiceConfig       `json:"management_service"`
+	MonitoringConfig        Monitoring                    `json:"monitoring_config"`
+	SwaggerConfig           SwaggerConfig                 `json:"swagger_config"`
+	SegmenterConfig         map[string]interface{}        `json:"segmenter_config"`
+}
+
+type AssignedTreatmentLoggerConfig struct {
```
krithika369, line 79: I imagine that eventually the plugin will have a direct dependency on the treatment-service modules. Given this, shall we reuse the structs from `treatment-service/config/config.go` here too, instead of redefining them ?

krithika369 (commented):
On the whole LGTM. Thanks for the changes! Left some comments / questions. Will have another look once they are reviewed and addressed.

plugins/turing/config/config.go
deadlycoconuts: That's true, I was planning to remove `RunnerDefaults` once I've refactored the plugin's experiment runner, so nope we don't need it anymore once that happens but for now it's still here because the experiment runner's still serving treatments with the endpoint. But yep, once the refactoring has taken place we can probably make the change as you have suggested, 'remove [the] struct and simply place the Timeout value under TreatmentServicePluginConfig'.

api/experiments.yaml
```diff
@@ -7,6 +7,17 @@ servers:
   - url: /v1
 tags: []
 paths:
+  /treatment-service-plugin-config:
```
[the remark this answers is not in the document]
deadlycoconuts: That's true, I think I'll just name the endpoint `/treatment-service-config` then; I was thinking about making the endpoint name reflect that the config served isn't a complete set of configs to run the treatment service but `/treatment-service-partial-config` might sound a little confusing 🤔

api/experiments.yaml
```diff
@@ -7,6 +7,17 @@ servers:
...
+    get:
+      operationId: GetTreatmentServicePluginConfig
+      tags:
+        - configuration
+      summary: retrieves configuration used in the management service that is also used in the treatment service plugin
```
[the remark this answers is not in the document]
deadlycoconuts: Okay I've renamed the description/summary for this struct to `retrieves treatment service configuration driven by the management service`.

api/schema.yaml
```diff
@@ -1,5 +1,57 @@
...
+    PubSub:
+      type: object
+      properties:
+        project:
+          type: string
+          description: Project name of the PubSub subscription
+        topic_name:
+          type: string
+          description: Topic name of the PubSub subscription
+    NewRelicConfig:
```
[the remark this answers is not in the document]
deadlycoconuts: Hmm I think we can move these configs to the Turing app. In any case, I believe the NewRelic/Sentry config contains stuff like the  License / DSN which we shouldn't be passing over the management service endpoint. There's also the other point which you made about the NewRelic App name / Sentry App label which are different from what the management service config contains.

plugins/turing/config/config.go
deadlycoconuts: Yep I've replaced all these structs entirely with those from the treatment service. Can't remember why I copied them over manually but I remember having done it for some reason 😵‍💫

deadlycoconuts:
Oh if you were wondering why there's a huge change in all the go mod files, I was bumping up all the common dependencies across the various parts of the repo. Brought that forward into this PR since we're beginning to import packages from the treatment service into the plugin (the treatment service config).

api/schema.yaml
```diff
@@ -1,5 +1,57 @@
...
+          description: Boolean indicating if Sentry is enabled
+        labels:
+          type: object
+          properties:
+    SegmenterConfig:
+      type: object
+      properties:
+        s2_ids:
+          $ref: '#/components/schemas/S2Ids'
+    S2Ids:
```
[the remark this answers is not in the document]
krithika369: @deadlycoconuts, addressing this comment is still WIP, yeah?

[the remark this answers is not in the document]
deadlycoconuts: Oh wait it's done! The `SegmenterConfig` schema definition has already been reduced to `type:object`; I just forgot to remove the S2Ids schema definition (will be removing this now).

management-service/services/configuration_service.go
```diff
@@ -0,0 +1,39 @@
...
+type configurationService struct {
+	treatmentServiceConfig schema.TreatmentServiceConfig
+}
+
+func NewConfigurationService(cfg *config.Config) ConfigurationService {
+	var segmenterConfig schema.SegmenterConfig
+	segmenterConfig = cfg.SegmenterConfig
+
+	// Iterates through all Sentry config labels to cast them as the type interface{}
+	sentryConfigLabels := make(map[string]interface{})
```
krithika369, line 21: We don't need this anymore, right?

krithika369 (approved):

krithika369:
Left a couple of small comments. The rest LGTM! Thanks for this PR, @deadlycoconuts. Please feel free to merge once you're done assessing the comments.

management-service/services/configuration_service.go
deadlycoconuts: Oops yes we don't need that anymore 😅 Thanks for catching this!

plugins/turing/config/config.go
```diff
@@ -31,20 +34,35 @@ type RemoteUI struct {
...
-	ProjectID         int        `json:"project_id" validate:"required"`
-	Passkey           string     `json:"passkey" validate:"required"`
-	Timeout           string     `json:"timeout" validate:"required"`
-	RequestParameters []Variable `json:"request_parameters" validate:"required,dive"`
+	Endpoint               string         `json:"endpoint" validate:"required"`
+	ProjectID              int            `json:"project_id" validate:"required"`
+	Passkey                string         `json:"passkey" validate:"required"`
+	Timeout                string         `json:"timeout" validate:"required"`
+	RequestParameters      []Variable     `json:"request_parameters" validate:"required,dive"`
+	TreatmentServiceConfig *config.Config `json:"treatment_service_config"`
```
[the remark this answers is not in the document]
deadlycoconuts: Hmm yep we should, although we haven't actually defined the validation tags for the treatment service's configs yet 😅 I'll add the `required` tag sparingly to fields certain fields for now, namely, `DebugConfig`, `DeploymentConfig`, `PubSub`, `ManagementService`, `SwaggerConfig` and `Port`. The other fields are either monitoring/logging configs that I'll add later or are optional.

deadlycoconuts:
Thanks for looking through this PR and for all the comments @krithika369 ! 🙏 I'll be merging this now! :)

PR #48, opened 2022-10-26
Bugfix: Navigation with react-router-dom

krithika369:
**What this PR does / why we need it**: This PR is a follow up from https://github.com/caraml-dev/xp/pull/45 which addresses the navigation regressions introduced by that PR.

Fixes the following issues:
* `ui/src/PrivateLayout.js`, `ui/src/AppRoutes.js` - When the XP app is accessed standalone (primarily during local development), the header bar's project change dropdown was not updating the UI states properly. This is because XP is not a valid first-level MLP app and therefore, `currentApp` would be undefined and cannot be used in `onProjectSelect` as with other apps like Turing. Directly navigating to `/xp` via `onProjectSelect` handler also results in incorrect states because of the underlying implementation of the [ProjectsContext](https://github.com/gojek/mlp/blob/main/ui/packages/lib/src/providers/project/context.js#L39) which requires the path to begin with `currentApp.href`. So, this is now handled by redirecting to `/projects/{pId}` (`currentApp.href` = "/") on project change which will subsequently redirect to `/xp` and then `/xp/projects/{pId}/experiments` via other redirects in `AppRoutes`.
* `ui/src/PrivateLayout.js` - Header's homepage should be empty (default "/") so that in the multi-app deployment, clicking the header logo can take us to the MLP landing page.
* `ui/src/experiments/ExperimentsLandingPage.js` - Fixed missing `/*` pattern to match nested routes
* `ui/src/experiments/list/ListExperimentsTable.js` - Some differences in handling of trailing slash between react router v5 and v6 require special handling for the Link navigation.

### DIFF
--- a/ui/src/experiments/ExperimentsLandingPage.js
+++ b/ui/src/experiments/ExperimentsLandingPage.js
@@ -68,7 +68,7 @@ const ExperimentsLandingPage = () => {
         which results in incorrect /experiments/experiments prefix.
          */}
           <Route
-            path="experiments"
+            path="experiments/*"
             element={<Navigate to={location.pathname.replace("/experiments/experiments", "/experiments")}
               replace={true} />} />
         </Routes>
--- a/ui/src/experiments/list/ListExperimentsTable.js
+++ b/ui/src/experiments/list/ListExperimentsTable.js
@@ -1,6 +1,8 @@
 import React from "react";
 
 import { EuiHealth, EuiLink, EuiText } from "@elastic/eui";
+import { useLocation } from "react-router-dom";
+import urlJoin from "proper-url-join";
 
 import { useConfig } from "config";
 import { BasicTable } from "components/table/BasicTable";
@@ -17,6 +19,7 @@ const ListExperimentsTable = ({
   onRowClick,
 }) => {
   const { appConfig } = useConfig();
+  const location = useLocation();
   const tableColumns = [
     {
       field: "id",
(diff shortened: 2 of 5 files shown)


ui/src/experiments/ExperimentsLandingPage.js
```diff
@@ -68,7 +68,7 @@ const ExperimentsLandingPage = () => {
         which results in incorrect /experiments/experiments prefix.
          */}
           <Route
-            path="experiments"
```
terryyylim, line 71: Just curious, how come removal of this didn't show any error previously? https://github.com/caraml-dev/xp/pull/45/files#diff-557155280a10fe2c6a10a1b299138e6f690220fb5ecffb8546ce3aa459a57287L71

ui/src/experiments/list/ListExperimentsTable.js
```diff
@@ -88,11 +91,12 @@ const ListExperimentsTable = ({
       width: "5%",
       render: (item) => {
         return (
+          // We need to do urlJoin to correctly handle trailing slash when used as a remote component
           <EuiLink
             onClick={(e) => {
               e.stopPropagation();
             }}
-            href={item.id}
+            href={urlJoin(location.pathname, item.id)}
```
terryyylim, line 99: Likewise for this - https://github.com/caraml-dev/xp/pull/45/files#diff-bad9138e8905bf7269b6399a5d022254ef273007f923a95f8e08ca199148ac2aL96.

terryyylim (approved):

ui/src/experiments/ExperimentsLandingPage.js
krithika369: The problem is noticed when we are using the context menu to switch from one nested section to Experiments (eg: /segments/{id} to /experiments). I missed this check when testing locally previously.

ui/src/experiments/list/ListExperimentsTable.js
krithika369: The previous change worked from the xp app, not when accessed as a remote component. Something that I hadn't checked when testing locally.

PR #49, opened 2022-10-28
add protocol for prediction_context display

leonlnj:
UI enhancement to support UPI routers in Turing. Most code changes are due to formatting. 

The only main change is from / `xp/ui/src/turing/components/form/EditExperimentEngineConfig.js` whereby protocol is passed from Turing, all the way to `xp/ui/src/turing/components/form/variables_config/components/FieldSourceFormLabel.js`. 

The value of `Prediction Context` stays the same as payload, as XP only supports HTTP, it is the Turing router that parses the UPI proto and send it to XP. This is mainly for visual aid to be consistency with traffic rules which shows `Prediction Context` too.

If no protocol is passed in, the original `Payload` will be shown.

image

### DIFF
--- a/ui/src/turing/components/form/variables_config/components/FieldSourceFormLabel.js
+++ b/ui/src/turing/components/form/variables_config/components/FieldSourceFormLabel.js
@@ -10,23 +10,30 @@ import { flattenPanelTree, useToggle } from "@gojek/mlp-ui";
 
 import "./FieldSourceFormLabel.scss";
 
-const fieldSourceOptions = [
-  {
-    value: "none",
-    inputDisplay: "None",
-  },
-  {
-    value: "header",
-    inputDisplay: "Header",
-  },
-  {
-    value: "payload",
-    inputDisplay: "Payload",
-  },
-];
-
-export const FieldSourceFormLabel = ({ value, onChange, readOnly }) => {
+export const FieldSourceFormLabel = ({
+  value,
+  protocol,
+  onChange,
+  readOnly,
+}) => {
   const [isOpen, togglePopover] = useToggle();
+
+  const fieldSourceOptions = [
+    {
+      value: "none",
+      inputDisplay: "None",
+    },
+    {
+      value: "header",
+      inputDisplay: "Header",
+    },
+    {
+      value: "payload",
+      // Display is change to Prediction Context to be consistent with Turing Traffic rule
+      // backend value stays the same as payload, because XP is not supporting gRPC
+      inputDisplay: protocol === "UPI_V1" ? "Prediction Context" : "Payload",
+    },
+  ];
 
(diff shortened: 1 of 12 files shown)


pradithya (approved):
> whereby protocol is passed from Turing, all the way to xp/ui/src/turing/components/form/variables_config/components/FieldSourceFormLabel.js.

I wonder if using context will be good here. It seems a bit overkill too. 

Other than that, LGTM!

terryyylim (approved):
LGTM, thanks @leonlnj ! Curious - how did you generate these formatting changes? Are we not capturing them correctly in pre-commit linting step?

ui/src/turing/components/form/variables_config/components/FieldSourceFormLabel.js
```diff
@@ -10,24 +10,31 @@ import { flattenPanelTree, useToggle } from "@gojek/mlp-ui";
...
+    },
+    {
+      value: "header",
+      inputDisplay: "Header",
+    },
+    {
+      value: "payload",
+      // Display is change to Prediction Context to be consistent with Turing Traffic rule
+      // backend value stays the same as payload, because XP is not supporting gRPC
+      inputDisplay: protocol === "UPI_V1" ? "Prediction Context" : "Payload",
```
krithika369, line 34: We'd also need to make this change for the details view, right?

https://github.com/caraml-dev/xp/blob/main/ui/src/turing/components/configuration/variables_config/VariablesConfigGroup.js#L28

Considering this, it would be nice to extract this into a utility function and place it under a common file in `turing/components/`.

krithika369 (approved):
Left one comment. The rest LGTM. Thanks!

ui/src/turing/components/form/variables_config/components/FieldSourceFormLabel.js
leonlnj: updated 
image

leonlnj:
> Curious - how did you generate these formatting changes? Are we not capturing them correctly in pre-commit linting step?

I was using vs code extension prettier which format on save (if it doesn't work i write prettier cli manually)

ui/src/turing/components/utils/helper.js
```diff
@@ -0,0 +1,11 @@
+export const mapProtocolLabel = (protocol, value) => {
+  if (protocol === "UPI_V1" && value === "payload") {
+    return "prediction context";
+  }
+
+  return value;
+};
+
+export const capitalizeFirstLetter = (string) => {
+  return string.replace(/(^\w|\s\w)/g, (m) => m.toUpperCase());
```
krithika369, line 10: Rather than the custom function, we could use this: https://lodash.com/docs#startCase

PR #52, opened 2022-11-22
Refactor treatment service into plugin

deadlycoconuts:
This PR is the second of a series of changes intended to offer the Treatment Service as a [Turing](https://github.com/caraml-dev/turing/) plugin, in addition to it being deployable as a standalone service, which is the currently the only way to deploy it. 

In PR #47, the plugin was refactored to be initialised with a number of additional configuration that the Treatment Service uses, though its functionality has not been modified yet. This PR introduces changes to the functionality of the plugin so that it no longer contacts the central Treatment Service (the standalone deployment) and instead initialises the necessary services that it requires to serve treatments itself (as if it were a separate Treatment Service itself).

**Modifications Overview**:
This PR heavily refactors the `GetTreatmentForRequest` interface method that's implemented by the `experimentRunner` struct. It shares very similar behaviour with (and is largely inspired by) the treatment controller method `FetchTreatment` of the central Treatment Service. In order to initialise the necessary services to assign treatments, the experiment runner is initialised with an app context, which is performed in the refactored `NewExperimentRunner`.

In addition to the main changes above, additional refactoring is done to the initialisation methods for the internal storage, BigQuery treatment logger and PubSub subscriber within the Treatment Service in order to allow the Google clients used in these services to be authenticated with a service account token stored in an alternate environment variable `EXP_GOOGLE_APPLICATION_CREDENTIALS` instead of the default `GOOGLE_APPLICATION_CREDENTIALS` variable (refer to https://github.com/caraml-dev/turing/pull/287 for more details). 

This is done so that when these services are initialised as part of a plugin, they would be able to utilise the alternate file path to locate the service account token as opposed to using the default environment variable, which currently indicates the location of the service account token of the Turing Router. Should the environment variable `EXP_GOOGLE_APPLICATION_CREDENTIALS` not be specified, these services would resort to reading the file path from `GOOGLE_APPLICATION_CREDENTIALS` - this is currently happening when these services get initialised as part of the central standalone Treatment Service.

Last but not least, some refactoring was done to make the retrieval of project settings in the `init` step of the Treatment Service's local storage simpler and more reader-friendly.


plugins/turing/config/config.go
```diff
@@ -46,16 +46,14 @@ type ExperimentManagerConfig struct {
...
 	ProjectID              int            `json:"project_id" validate:"required"`
-	Passkey                string         `json:"passkey" validate:"required"`
 	Timeout                string         `json:"timeout" validate:"required"`
 	RequestParameters      []Variable     `json:"request_parameters" validate:"required,dive"`
 	TreatmentServiceConfig *config.Config `json:"treatment_service_config" validate:"required,dive"`
 }
 
 type TreatmentServicePluginConfig struct {
-	Port                 int      `json:"port" default:"8080"`
-	ProjectIds           []string `json:"project_ids" default:""`
```
deadlycoconuts, line 57: Removed `ProjectIds` as this information can be retrieved directly by the json config passed to `GetExperimentRunnerConfig` (the Turing API indirectly calls this when deploying a new Turing Router).

plugins/turing/runner/experiment_runner.go
```diff
@@ -44,67 +49,101 @@ func (er *experimentRunner) GetTreatmentForRequest(
...
 	// Get the request parameters for the current request
-	requestParams := map[string]string{}
+	requestParams := map[string]interface{}{}
 	for _, param := range er.parameters {
 		if param.FieldSource == "none" || param.Field == "" {
 			// Parameter not configured
 			continue
 		}
-		val, err := request.GetValueFromRequest(reqHeader, body, request.FieldSource(param.FieldSource), param.Field)
+		val, err := request.GetValueFromHTTPRequest(reqHeader, body, request.FieldSource(param.FieldSource), param.Field)
```
deadlycoconuts, line 146: Updated this when I bumped up the dependency on `github.com/caraml-dev/turing/engines/experiment`.

plugins/turing/manager/experiment_manager_test.go
```diff
@@ -179,19 +177,9 @@ func TestGetExperimentRunnerConfig(t *testing.T) {
 		},
 	}
 
-	// Patch method to get passkey
+	// Patch method to get treatment service config
 	// TODO: Generate mock client and use it here instead of patching
 	em := &experimentManager{RunnerDefaults: _config.RunnerDefaults{Endpoint: "test-endpoint", Timeout: "12s"}}
-	monkey.PatchInstanceMethod(
-		reflect.TypeOf(em),
-		"GetProject",
```
krithika369, line 187: Do we need the `GetProject` function on the experiment manager anymore? Can we remove it?

plugins/turing/runner/experiment_runner.go
```diff
@@ -148,8 +193,11 @@ func NewExperimentRunner(jsonCfg json.RawMessage) (runner.ExperimentRunner, erro
 	r := &experimentRunner{
 		httpClient: client,
 		projectID:  config.ProjectID,
-		passkey:    config.Passkey,
 		parameters: config.RequestParameters,
+		appContext: appCtx,
 	}
+
+	errChannel := make(chan error, 1)
```
krithika369, line 200: Can we leave a //TODO note here to handle the error later from `errChannel` later on?

treatment-service/models/storage.go
```diff
@@ -5,22 +5,24 @@ import (
...
 	managementClient "github.com/caraml-dev/xp/clients/management"
 	"github.com/caraml-dev/xp/common/api/schema"
 	"github.com/caraml-dev/xp/common/pubsub"
 	_segmenters "github.com/caraml-dev/xp/common/segmenters"
+	"github.com/golang-collections/collections/set"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
 )
 
+const ExpGoogleApplicationCredentials = "EXP_GOOGLE_APPLICATION_CREDENTIALS"
```
krithika369, line 24: This is reasonable, but it's a bit strange to let this common code be driven by Turing's choice of env var for `EXP_GOOGLE...`. I think it will be better to make it a config (perhaps in the [DeploymentConfig](https://github.com/caraml-dev/xp/blob/main/treatment-service/config/config.go#L72) struct, as `GoogleApplicationCredentialsEnvVar` - this config is applicable to PubSub / Management API calls now, later on to BQ logging of the treatment generation as well), much like other config values that can vary across deployments. Its default value can be unset. This will still match the implementation you have in `treatment-service/services/subscription.go`.

Wdyt?

krithika369 (approved):
Left some small comments - the PR looks great, thanks for this and iteratively testing and resolving the issues. Please feel free to merge once you've reviewed / addressed the comments.

treatment-service/models/storage.go
deadlycoconuts: That makes sense; I've just refactored all the implementations in the relevant services to read off the filepath from the `GoogleApplicationCredentialsEnvVar` config stored in the `DeploymentConfig`.

PR #53, opened 2022-12-01
Bugfix: Fix retrieval of navigate function

deadlycoconuts:
This PR fixes a tiny bug with the UI when a user is led to the landing page with a prompt to create new project settings for a project that has yet to be set up.

Upon clicking on the link in the prompt, the user would not get redirected correctly and nothing happens.

Screenshot 2022-12-01 at 5 51 46 PM

This PR fixes the way in which the `navigate` function from the `react-router-dom` package gets retrieved.


terryyylim (approved):

deadlycoconuts:
Thanks for the quick approval! 🚀

PR #54, opened 2022-12-15
Add metrics logging to plugin

deadlycoconuts:
This PR is the last of a series of changes to offer the Treatment Service as a [Turing](https://github.com/caraml-dev/turing/) plugin, in addition to it being deployable as a standalone service, which is the currently the only way to deploy it.

In PR https://github.com/caraml-dev/xp/pull/52, the plugin was refactored to initialise the necessary services that it requires to serve treatments itself (as if it were a separate Treatment Service). This PR introduces changes dependent on several Turing updates (see https://github.com/caraml-dev/turing/pull/302) in order to allow the plugin to perform metrics logging using the Turing Router's metrics collector.

This PR makes changes to the plugin by introducing additional steps to set the metrics collector passed to it from the Turing Router as the global metrics collector, and to register its own metrics on it. Minor refactoring has been done to the metrics service so as to make certain logging methods more reusable by the plugin, instead of importing the controller-layer level helper methods (`treatment-service/controller/treatment.go`).


plugins/turing/go.mod
```diff
@@ -123,7 +127,10 @@ require (
 )
 
 replace (
+	github.com/caraml-dev/turing/engines/experiment => ../../../turing/engines/experiment
```
krithika369, line 130: I know this PR is WIP but this needs to be removed. :)

plugins/turing/runner/experiment_runner.go
```diff
@@ -109,13 +134,60 @@ func (er *experimentRunner) GetTreatmentForRequest(
...
+			Name:        string(instrumentation.ExperimentLookupDurationMs),
+			Type:        routerMetrics.HistogramMetricType,
+			Description: instrumentation.ExperimentLookupDurationMsHelpString,
+			Buckets:     instrumentation.RequestLatencyBuckets,
+			Labels:      instrumentation.ExperimentLookupDurationMsLabels,
+		},
+		{
+			Name:        string(instrumentation.FetchTreatmentRequestCount),
+			Type:        routerMetrics.CounterMetricType,
+			Description: "Counter for no. of Fetch Treatment requests with matching experiments",
```
krithika369, line 169: I guess this description and the next one can also be re-used from the `instrumentation` package?

treatment-service/services/metric_service.go
```diff
@@ -7,34 +7,43 @@ import (
...
 type metricService struct {
 	Kind         config.MetricSinkKind
 	LocalStorage *models.LocalStorage
+	MetricLabels []string
 }
 
 func NewMetricService(cfg config.Monitoring, localStorage *models.LocalStorage) (MetricService, error) {
 	switch cfg.Kind {
-	case config.NoopMetricSink:
+	case config.NoopMetricSink, config.RPCMetricSink:
```
krithika369, line 46: The implementation for the RPC sink is still WIP, correct?

plugins/turing/go.mod
deadlycoconuts: Whoops thanks for highlighting that! :)

plugins/turing/runner/experiment_runner.go
deadlycoconuts: Yep that's true, I've updated it to reuse the description from that package.

treatment-service/services/metric_service.go
deadlycoconuts: Actually no, the `RPCMetricSink` is just a metric service type that informs the metric service to not initialise a new prometheus metrics collector, but to still use the logging methods `LogRequestCount` and `LogLatencyHistogram` as normal.

treatment-service/config/config_test.go
```diff
@@ -19,7 +19,7 @@ func TestDefaultConfigs(t *testing.T) {
 			AllowedOrigins:   []string{"*"},
 			OpenAPISpecsPath: ".",
 		},
-		ProjectIds: nil,
+		ProjectIds: []string{},
```
deadlycoconuts, line 22: This change is made because the default config object created by viper now specifies an empty array instead of nil by default when that field is unspecified.

treatment-service/instrumentation/prometheus.go
```diff
@@ -18,36 +18,60 @@ const (
...
+	// FetchTreatmentRequestCountHelpString is the help string of the FetchTreatmentRequestCount metric
+	FetchTreatmentRequestCountHelpString string = "Counter for no. of Fetch Treatment requests with matching experiments"
+	// NoMatchingExperimentRequestCountHelpString is the help string of the NoMatchingExperimentRequestCount metric
+	NoMatchingExperimentRequestCountHelpString string = "Counter for no. of Fetch Treatment requests with no matching experiments"
 )
 
-// requestLatencyBuckets defines the buckets used in the custom Histogram metrics
-var requestLatencyBuckets = []float64{
+// RequestLatencyBuckets defines the buckets used in the custom Histogram metrics
+var RequestLatencyBuckets = []float64{
```
deadlycoconuts, line 32: Global definitions such as these are made public so that xp's plugin package can import them.

treatment-service/controller/treatment.go
```diff
@@ -201,41 +201,7 @@ func (t TreatmentController) FetchTreatment(w http.ResponseWriter, r *http.Reque
 	Ok(w, response, &requestId)
 }
 
-func (t TreatmentController) logFetchTreatmentMetrics(
```
deadlycoconuts, line 204: This method has been refactored as a metric service method so that it can be reused by xp's plugin.

treatment-service/services/metric_service.go
```diff
@@ -48,16 +57,22 @@ func NewMetricService(cfg config.Monitoring, localStorage *models.LocalStorage)
 	svc := &metricService{
 		Kind:         cfg.Kind,
 		LocalStorage: localStorage,
+		MetricLabels: cfg.MetricLabels,
```
deadlycoconuts, line 60: The `metricService` struct now keeps a copy of the metric labels found in the config so that the plugin (which does not store a copy of the monitoring configs, unlike the global treatment service) gets to use them for logging.

treatment-service/services/metric_service.go
krithika369: Ok I think I got the difference between my assumption and the implementation now, after referring to the Turing PR: https://github.com/caraml-dev/turing/pull/302

Instead of the `RegisterMetricsCollector` call from Turing which effectively "sets" the collector on XP, can we have XP "get" the collector from Turing under the case `RPCMetricSink` here ? This would be a more natural flow, IMO. Wdyt?

deadlycoconuts: Hmm what you suggested is indeed more intuitive but I left the current implementation this way because of the order in which the metrics collector is received with respect to the creation of the experiment runner.

Currently, the app context (which includes the metric service) gets created during the creation of a new experiment runner (this happens in `NewExperimentRunner`). After this call returns, the newly created experiment runner then gets to 'set' the collector when the Turing Router calls `RegisterMetricsCollector`. 

In order to have the metric service set the collector when the metric service is being initialised (as part of the app context initialisation), we would need to somehow pass the metrics collector to whichever process is calling `NewExperimentRunner`, even before the experiment runner has been created. I'm not 100% sure right now how this would work but I can look into it.

krithika369 (approved):
Other than [this pending comment](https://github.com/caraml-dev/xp/pull/54#discussion_r1060315758), LGTM. Thanks for the changes! Please feel free to merge when done reviewing the comment. 🎉

treatment-service/services/metric_service.go
```diff
@@ -7,34 +7,43 @@ import (
...
 	LogRequestCount(labels map[string]string, loggingMetric metrics.MetricName)
 
 	// GetProjectNameLabel retrieves only project name as labels
 	GetProjectNameLabel(projectId models.ProjectId) map[string]string
+	GetMetricLabels() []string
 	// GetLabels retrieves labels with flag to filter for segmenters
 	GetLabels(projectId models.ProjectId, treatment schema.SelectedTreatment, statusCode int,
-		metricLabels []string, requestFilter map[string][]*_segmenters.SegmenterValue, withSegmenters bool) map[string]string
+		requestFilter map[string][]*_segmenters.SegmenterValue, withSegmenters bool) map[string]string
+	ReplacePrometheusCollector(collector metrics.Collector)
```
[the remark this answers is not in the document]
deadlycoconuts: Whoops, thanks for picking that out! Was playing around with random names then and didn't think too deeply into them 😅

deadlycoconuts:
Thanks a lot for the review! I'll be merging this! :)

PR #55, opened 2022-12-16
Init empty array for List experiments response

krithika369:
**What this PR does / why we need it**: This is a minor bugfix for the List Experiments API when 0 records are selected from the DB. We subsequently iterate over the experiments and copy only the required fields (specified by the field selector, if set), to a new array. However, this array was previously uninitialised, leading to `"data": null` as opposed to `"data": []` in the response. (This is implemented correctly for other entities like Segments and Treatments already.)


terryyylim (approved):

krithika369:
Thanks for the quick review @terryyylim !

PR #56, opened 2023-01-10
Missing randomization key in Fetch Treatment request should be treated as a bad request

krithika369:

The randomization key is only required if an A/B experiment is selected by the Fetch Treatment request handler. However, if the randomization key is missing in the original request, treatment assignment cannot be done. This was previously being returned as a 5XX error (which wrongly indicates a problem with the server). This PR corrects this scenario to be dealt with as a 4XX error (Bad User Request).


deadlycoconuts (approved):
LGTM! 🚀 Thanks for the quick fix!

krithika369:
Thanks for the quick review @deadlycoconuts ! Merging this.

PR #58, opened 2023-01-17
Remove redundant configs from experiment runner

deadlycoconuts:
After the plugin has been refactored to incorporate the Treatment Service, it has been found that there were a couple of redundant fields that were left hanging in the experiment manager/runner configs that were passed around. This PR simply serves to remove those fields from the said configs:
- `ExperimentManagerConfig.RunnerDefaults` - since the plugin no longer communicates with a central Treatment Service, there is no need for an API endpoint or timeout value
- `ExperimentRunnerConfig.Endpoint` -  same as above
- `ExperimentRunnerConfig.Timeout` - same as above
- `ExperimentRunnerConfig.ProjectID` - this value is a duplicate of the value found in `ExperimentRunnerConfig.TreatmentServiceConfig.ProjectIds`

### DIFF
--- a/plugins/turing/runner/experiment_runner.go
+++ b/plugins/turing/runner/experiment_runner.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/caraml-dev/turing/engines/experiment/log"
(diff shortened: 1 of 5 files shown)


plugins/turing/runner/experiment_runner.go
```diff
@@ -37,8 +37,7 @@ func init() {
 
 // experimentRunner implements runner.ExperimentRunner
 type experimentRunner struct {
-	httpClient *xpclient.ClientWithResponses
```
deadlycoconuts, line 40: There is no need for this client since the experiment runner no longer contacts the central Treatment Service.

krithika369 (approved):
Thanks for cleaning this up, @deadlycoconuts ! LGTM.

PR #59, opened 2023-01-19
Reduce treatment service memory usage

deadlycoconuts:
This PR introduces some minor refactoring that reduces the memory usage of the Treatment Service by removing the redundancy in storing segmenter values twice in the `ExperimentIndex` struct:
- in the various sets, i.e. `ExperimentIndex.stringSets`, `ExperimentIndex.intSets`, `ExperimentIndex.realSets` 
- in the field `ExperimentIndex.Experiment.Segments`

Given that the Treatment Service only uses the sets of an `ExperimentIndex` to determine a matching experiment, there is no need to store the original copy within `ExperimentIndex.Experiment.Segments`. While this duplication might seem minor and negligible, we have observed instances whereby experiments with a huge number of segmenter values cause a huge increase in memory usage by the Treatment Service due to the duplication of the data stored in memory.

This PR simply removes the values in `ExperimentIndex.Experiment.Segments` once they have been used to initialise the various sets, to reduce the overall memory footprint of the Treatment Service. Additional changes to existing functions are also made to accommodate the removal of that field.

### DIFF
--- a/common/utils/typeconverter.go
+++ b/common/utils/typeconverter.go
@@ -13,11 +13,11 @@ func StringSliceToListSegmenterValue(values *[]string) *_segmenters.ListSegmente
 	if values == nil {
 		return nil
 	}
-	segmenterValues := []*_segmenters.SegmenterValue{}
-	for _, val := range *values {
-		segmenterValues = append(segmenterValues, &_segmenters.SegmenterValue{
-			Value: &_segmenters.SegmenterValue_String_{String_: val},
-		})
+	segmenterValues := make([]*_segmenters.SegmenterValue, len(*values))
+	for i := 0; i < len(*values); i++ {
+		segmenterValues[i] = &_segmenters.SegmenterValue{
+			Value: &_segmenters.SegmenterValue_String_{String_: (*values)[i]},
+		}
 	}
 	return &_segmenters.ListSegmenterValue{Values: segmenterValues}
 }
@@ -26,9 +26,11 @@ func BoolSliceToListSegmenterValue(values *[]bool) *_segmenters.ListSegmenterVal
 	if values == nil {
 		return nil
 	}
-	segmenterValues := []*_segmenters.SegmenterValue{}
-	for _, val := range *values {
-		segmenterValues = append(segmenterValues, &_segmenters.SegmenterValue{Value: &_segmenters.SegmenterValue_Bool{Bool: val}})
+	segmenterValues := make([]*_segmenters.SegmenterValue, len(*values))
+	for i := 0; i < len(*values); i++ {
+		segmenterValues[i] = &_segmenters.SegmenterValue{
+			Value: &_segmenters.SegmenterValue_Bool{Bool: (*values)[i]},
+		}
 	}
 	return &_segmenters.ListSegmenterValue{Values: segmenterValues}
 }
@@ -37,11 +39,11 @@ func Int64ListToListSegmenterValue(values *[]int64) *_segmenters.ListSegmenterVa
 	if values == nil {
 		return nil
 	}
-	segmenterValues := []*_segmenters.SegmenterValue{}
-	for _, val := range *values {
-		segmenterValues = append(segmenterValues, &_segmenters.SegmenterValue{
-			Value: &_segmenters.SegmenterValue_Integer{Integer: val},
-		})
+	segmenterValues := make([]*_segmenters.SegmenterValue, len(*values))
+	for i := 0; i < len(*values); i++ {
+		segmenterValues[i] = &_segmenters.SegmenterValue{
+			Value: &_segmenters.SegmenterValue_Integer{Integer: (*values)[i]},
+		}
 	}
 	return &_segmenters.ListSegmenterValue{Values: segmenterValues}
 }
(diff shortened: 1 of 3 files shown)


common/utils/typeconverter.go
```diff
@@ -13,11 +13,11 @@ func StringSliceToListSegmenterValue(values *[]string) *_segmenters.ListSegmente
...
-	segmenterValues := []*_segmenters.SegmenterValue{}
-	for _, val := range *values {
-		segmenterValues = append(segmenterValues, &_segmenters.SegmenterValue{
-			Value: &_segmenters.SegmenterValue_String_{String_: val},
-		})
+	segmenterValues := make([]*_segmenters.SegmenterValue, len(*values))
+	for i := 0; i < len(*values); i++ {
+		segmenterValues[i] = &_segmenters.SegmenterValue{
+			Value: &_segmenters.SegmenterValue_String_{String_: (*values)[i]},
+		}
```
deadlycoconuts, line 20: This is a series of minor changes to initialise a fixed-length slice instead of a dynamic length slice to store the segmenter values so as to prevent repeated allocation of memory to copy-paste the slice values when `append`-ing.

treatment-service/models/storage.go
```diff
@@ -78,6 +78,7 @@ type ExperimentIndex struct {
 	stringSets map[string]*set.Set
 	intSets    map[string]*set.Set
 	realSets   map[string]*set.Set
+	boolFlags  map[string]bool
```
deadlycoconuts, line 81: Added this new field to the `ExperimentIndex` struct to store boolean segmenter values.

treatment-service/models/storage.go
```diff
@@ -238,6 +234,22 @@ func (i *ExperimentIndex) isActive() bool {
...
+	}
+	if set, exists := i.realSets[segmentName]; !exists || set.Len() == 0 {
+		return true
+	}
+	if _, exists := i.boolFlags[segmentName]; !exists {
+		return true
+	}
+	return false
+}
+
```
deadlycoconuts, line 252: This new function is basically performing the same purpose as 

```go
if _, exists := i.Experiment.Segments[segmentName]; !exists ||
			len(i.Experiment.Segments[segmentName].GetValues()) == 0 {
// ...
}
```
but has been refactored into a separate helper function and also uses the values within the sets of `ExperimentIndex` instead of the values stored in `ExperimentIndex.Experiment.Segments` (which would have been deleted).

treatment-service/models/storage_test.go
```diff
@@ -672,7 +675,7 @@ func TestExperimentIndexMatchSegment(t *testing.T) {
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
 			got := experimentIndex.matchSegment(tt.args.segmentName, tt.args.value)
-			assert.Equal(t, got, tt.want)
+			assert.Equal(t, tt.want, got)
```
deadlycoconuts, line 678: Flipped this around because `assert.Equal` was used incorrectly.

treatment-service/models/storage.go
krithika369: This needs to be a `set.Set` as well. It's possible to have a bool segment as `{true, false}` if it applies to both values.

treatment-service/models/storage.go
```diff
@@ -384,15 +397,22 @@ func NewExperimentIndex(experiment *pubsub.Experiment) *ExperimentIndex {
...
 				realSets[key].Insert(val.GetReal())
+			case *_segmenters.SegmenterValue_Bool:
+				boolFlags[key] = val.GetBool()
 			}
 		}
 	}
 
+	// Delete all segments since they have already been converted to the various sets stored in ExperimentIndex,
+	// and are no longer used by the Treatment Service
+	experiment.Segments = nil
```
krithika369, line 408: Could you leave a `TODO` note here with the long term idea of letting the `ExperimentIndex` store only the relevant data using appropriate structs rather than attempting to reuse this pubsub message type and deleting the redundant data from it?

treatment-service/models/storage.go
```diff
@@ -238,6 +234,22 @@ func (i *ExperimentIndex) isActive() bool {
 	return (i.StartTime.Before(time.Now()) || i.StartTime.Equal(time.Now())) && i.EndTime.After(time.Now())
 }
 
+func (i *ExperimentIndex) checkSegmentHasWeakMatch(segmentName string) bool {
+	if set, exists := i.stringSets[segmentName]; !exists || set.Len() == 0 {
```
krithika369, line 238: By this logic, int / real / bool segments will always return `true`. We should reverse the checks instead, like:

```golang
    if set, exists := i.stringSets[segmentName]; exists && set.Len() > 0 {
        return false
    }
    if ...
    return true
```

Better yet:

```golang
    if set, exists := i.stringSets[segmentName]; exists {
        if set.Len() > 0 {
                return false
        }
    } else if set, exists := i.intSets[segmentName]; exists {
        ...
    }
    ...
    return true
```

treatment-service/models/storage.go
deadlycoconuts: :O Really? But the original implementation of `matchFlagSetSegment` seems to suggest that the flag could only be one of `true` or `false`

krithika369: How? We are iterating over `range i.Experiment.Segments[segmentName].GetValues()`, no?

deadlycoconuts: Ohhh I see it! Oops my bad, I got carried away with the slightly different implementation of this function as compared to the other `match___SetSegment` functions and interpreted this differently 😅

treatment-service/models/storage.go
deadlycoconuts: Ooops logic failure there 😅 Thanks for catching this!

deadlycoconuts: Alright, just fixed this with your suggestion! :)

treatment-service/models/storage.go
deadlycoconuts: Added the `TODO` note just below that comment!

krithika369 (approved):
LGTM. Thanks for this fix! 🚀

deadlycoconuts:
Thanks a lot for the quick review! I'll be merging this now! 🚀

PR #61, opened 2023-02-01
Add logging to plugin fetch treatment method

deadlycoconuts:
This PR introduces the asynchronous logging method to the `GetTreatmentForRequest` method of the XP plugin that generates treatments. This functionality was previously left out from an earlier PR as a mistake, and this PR solely serves to fix that.

### DIFF
--- a/plugins/turing/go.mod
+++ b/plugins/turing/go.mod
@@ -19,6 +19,8 @@ require (
 	golang.org/x/oauth2 v0.0.0-20221014153046-6fdb5e3db783
 )
 
+require github.com/google/uuid v1.3.0
+
 require (
 	cloud.google.com/go v0.104.0 // indirect
 	cloud.google.com/go/bigquery v1.42.0 // indirect
@@ -61,7 +63,6 @@ require (
 	github.com/golang/groupcache v0.0.0-20210331224755-41bb18bfe9da // indirect
 	github.com/golang/protobuf v1.5.2 // indirect
 	github.com/google/go-cmp v0.5.9 // indirect
-	github.com/google/uuid v1.3.0 // indirect
 	github.com/googleapis/enterprise-certificate-proxy v0.2.0 // indirect
 	github.com/googleapis/gax-go/v2 v2.6.0 // indirect
 	github.com/grpc-ecosystem/grpc-gateway/v2 v2.11.2 // indirect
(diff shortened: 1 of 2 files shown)


plugins/turing/runner/experiment_runner.go
```diff
@@ -64,7 +66,9 @@ func (er *experimentRunner) GetTreatmentForRequest(
 	var treatment schema.SelectedTreatment
 	var switchbackWindowId *int64
 	var err error
+
 	statusCode := http.StatusBadRequest
+	filterParams := api.FetchTreatmentRequestBody{AdditionalProperties: requestParams}
```
deadlycoconuts, line 71: Created this new variable as it's being used twice.

plugins/turing/runner/experiment_runner.go
```diff
@@ -78,12 +82,65 @@ func (er *experimentRunner) GetTreatmentForRequest(
...
+				filterParams,
 				requestFilter,
 			)
 		}
 	}()
 
+	var lookupRequestFilters []models.SegmentFilter
+	var errorLog *monitoring.ErrorResponseLog
+
+	if er.appContext.AssignedTreatmentLogger != nil {
```
deadlycoconuts, line 94: This block is almost a perfect copy-paste from https://github.com/caraml-dev/xp/blob/f5eb2bd3c3ce301f392a1120232748a9255ab998/treatment-service/controller/treatment.go#L67.

plugins/turing/go.mod
```diff
@@ -19,6 +19,8 @@ require (
 	golang.org/x/oauth2 v0.0.0-20221014153046-6fdb5e3db783
 )
 
+require github.com/google/uuid v1.3.0
```
terryyylim, line 22: Is this intentional? - Why is this dependency not in the above `require` block?

deadlycoconuts: Hmm good point; not intentional but autogenerated from `go mod tidy`. Lemme poke this around 🤔

deadlycoconuts: Okay it seems like after manually shifting it and then running `go mod tidy`, the dependency stays in the block above now 👀

krithika369 (approved):
LGTM. Thanks for catching and fixing this!

deadlycoconuts:
Thanks for the quick review; merging this now! 🚀

PR #62, opened 2023-02-10
Bugfix: Make experiment runner return a 200 success code when no experiments are matched

deadlycoconuts:
There is a bug in the `GetTreatmentForRequest` method of the XP plugin that causes it to return a 400 error when no experiments are matched for the given request. This PR introduces a tiny fix to change the response code to 200 when this happens, just like how this case (no experiments matched) is handled by the standalone Treatment Service (see analogous line [here](https://github.com/caraml-dev/xp/blob/f5eb2bd3c3ce301f392a1120232748a9255ab998/treatment-service/controller/treatment.go#L161))


krithika369 (approved):
LGTM, thanks for tracing this!

PR #63, opened 2023-02-27
Update Treatment Validation doc with examples

krithika369:
**What this PR does / why we need it**: Adding some examples to the docs for the Go Template based treatment validation, since it's an area that we often get questions on.


deadlycoconuts (approved):
Looks good! Thanks for adding those 2 examples! 🚀

krithika369:
@deadlycoconuts thanks for the super quick review!

PR #64, opened 2023-03-02
Refactor Message Queue implementation

terryyylim:

In order for Management and Treatment Service to start without an actual Google PubSub, we need to remove the dependency on it, and this PR introduces a generic `MessageQueue` interface for doing so.

In addition, an enum has been introduced for the schema used by Treatment Service plugin, to support `noop` and `pubsub` message queue kinds.

Fixes #


management-service/services/message_queue_service.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"fmt"
+
+	_pubsub "github.com/caraml-dev/xp/common/pubsub"
+	"github.com/caraml-dev/xp/common/segmenters"
+	"github.com/caraml-dev/xp/management-service/config"
+)
+
+type MessageQueueService interface {
```
krithika369, line 11: Two thoughts:
* This interface makes the existing `PubSubPublisherService` redundant. We can remove it and its mock under `mocks`.
* Shall we make a nested `messagequeue` package under `services` and and put `pubsub_message_queue.go` (which is largely `management-service/services/pubsub_publisher_service.go`) and `noop_message_queue.go` as the interface implementations for this service under that? That will make the code organization better.

treatment-service/config/config.go
```diff
@@ -88,10 +88,28 @@ type Monitoring struct {
 	MetricLabels []string       `json:"metric_labels" default:""`
 }
 
+// MessageQueueKind describes the message queue for transmitting event updates to and fro Treatment Service
+type MessageQueueKind = string
```
krithika369, line 92: Since this is appearing in multiple places, shall we move these definitions to the `common` module, under a `messagequeue` package?

treatment-service/services/message_queue_service.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/caraml-dev/xp/treatment-service/config"
+	"github.com/caraml-dev/xp/treatment-service/models"
+)
+
+type MessageQueueService interface {
```
krithika369, line 11: Similar comment as with the management service. I think we should put the existing `subscription.go` and the new NoopMQ implementation in separate files, under a `messagequeue` package. Also the `ExperimentSubscriber ` interface and be removed.

krithika369 (commented):
Thanks for this refactoring @terryyylim ! The PR largely looks good. Left some comments on the organization of the code.

treatment-service/config/config.go
```diff
@@ -88,10 +88,28 @@ type Monitoring struct {
...
+	PubSubConfig PubSub
+}
+
 type PubSub struct {
-	Project              string `json:"project" default:"dev" validate:"required"`
-	TopicName            string `json:"topic_name" default:"xp-update" validate:"required"`
-	PubSubTimeoutSeconds int    `json:"pub_sub_timeout_seconds" default:"30" validate:"required"`
+	Project              string `json:"project" default:"dev"`
+	TopicName            string `json:"topic_name" default:"xp-update"`
+	PubSubTimeoutSeconds int    `json:"pub_sub_timeout_seconds" default:"30"`
```
deadlycoconuts, line 112: Why are we removing the validation tags? Can we have `MessageQueueConfig` have a pointer to the `PubSub`object instead so that field is nullable? i.e.
```go
type MessageQueueConfig struct {
	Kind MessageQueueKind `default:""`
	PubSubConfig *PubSub
}
```

treatment-service/server/server.go
```diff
@@ -98,6 +100,11 @@ func NewServer(configFiles []string) (*Server, error) {
 		mux.Handle("/schema.yaml", web.FileHandler(path.Join(cfg.SwaggerConfig.OpenAPISpecsPath, "schema.yaml"), false))
 	}
 
+	subscribe := false
+	if cfg.MessageQueueConfig.Kind != config.NoopMQ {
+		subscribe = true
+	}
```
deadlycoconuts, line 106: More like a nit than anything, but can we not use the `Server`'s `appContext` to determine whether the `MessageQueueService` is of the `NoopMQ` type or not? 😅 But I guess the type checking of that would make things more bulky and long-winded too :/

deadlycoconuts (approved):
LGTM! Thanks a lot for the extensive refactoring! Just left a couple of tiny comments! 🚀

management-service/config/config.go
```diff
@@ -53,6 +53,24 @@ type MLPConfig struct {
 	URL string
 }
 
+// MessageQueueKind describes the message queue for transmitting event updates to and fro Treatment Service
+type MessageQueueKind = string
+
+const (
+	// NoopMQ is a No-Op Message Queue
+	NoopMQ MessageQueueKind = ""
```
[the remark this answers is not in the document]
terryyylim: No reason, just that we don't need to specify anything in this case, makes the config look simpler.

management-service/config/config_test.go
```diff
@@ -38,9 +38,12 @@ func TestDefaultConfigs(t *testing.T) {
...
 		},
-		PubSubConfig: &PubSubConfig{
-			Project:   "dev",
-			TopicName: "xp-update",
+		MessageQueueConfig: &MessageQueueConfig{
+			Kind: "",
+			PubSubConfig: &PubSubConfig{
+				Project:   "dev",
+				TopicName: "xp-update",
+			},
```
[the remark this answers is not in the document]
terryyylim: Yes, those are default values.

management-service/config/example.yaml
```diff
@@ -29,9 +29,11 @@ DbConfig:
 MLPConfig:
   URL: http://localhost:8080/api/v1
 
-PubSubConfig:
-  Project: dev
-  TopicName: xp-update
+MessageQueueConfig:
+  # Kind: pubsub
```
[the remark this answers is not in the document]
terryyylim: Yes, default is noop.

tests/e2e/config/treatment-service.local.yaml
```diff
@@ -10,6 +10,9 @@ DeploymentConfig:
 NewRelicConfig:
   Enabled: false
 
+MessageQueueConfig:
+  Kind: pubsub
```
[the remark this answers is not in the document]
terryyylim: The default values are what's expected for e2e tests.

treatment-service/server/server.go
terryyylim: Hmm not sure what you mean, I'm using the config here right? 😮

treatment-service/config/config.go
terryyylim: 👍🏻 reverted!

krithika369 (approved):

treatment-service/server/server.go
deadlycoconuts: Ah I was just wondering if there was a way to avoid creating a new `subscribe` field in the `Server` struct and instead perform checks on the `appContext`'s `MessageQueueService` to see if it needs to subscribe to a message queue topic. But it's just a tiny comment, no worries about it.

PR #65, opened 2023-03-02
Set subscription expiry to 1 day

krithika369:
**What this PR does / why we need it**: PubSub subscriptions have a default expiry time of 31 days. From the PubSub docs:
> By default, subscriptions expire after 31 days of subscriber inactivity or if there are no updates made to the subscription.

This can result in a lot of clutter due to autoscaling (when some pods are removed, their subscriptions still live on GKE a while). So, this PR explicitly sets the [expiry](https://pkg.go.dev/cloud.google.com/go/pubsub#SubscriptionConfig) to 1 day.


deadlycoconuts (approved):
LGTM! Thanks for updating this!

PR #66, opened 2023-03-06
Support multi-stage builds for Management & Treatment Services Dockerfile

terryyylim:

When updating the helm chart for XP Management Service https://github.com/caraml-dev/helm-charts/pull/220, `Permission Denied` error was faced when trying to run the generated [binary](https://github.com/caraml-dev/helm-charts/actions/runs/4341758906/jobs/7581747169).

Subsequently, we're faced with `Exec format error` which points to our binary file being built on Ubuntu and to be used in an Alpine image. This PR adds multi-stage builds to build our app binaries using an Alpine image and copied over to the latest layer of our Docker image to solve the issue and also make the image leaner.

Fixes #

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,8 +35,7 @@ RUN addgroup -S ${XP_USER_GROUP} \
     && mkdir /app \
     && chown -R ${XP_USER}:${XP_USER_GROUP} /app
 
-COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/experiments.yaml /app/
-COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/schema.yaml /app/
+COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/*.yaml /app/
 COPY --from=api-builder --chown=${XP_USER}:${XP_USER_GROUP} /app/management-service/bin/* /app/
 COPY --from=api-builder --chown=${XP_USER}:${XP_USER_GROUP} /app/management-service/database /app/database/
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -35,8 +35,8 @@ RUN addgroup -S ${XP_USER_GROUP} \
     && mkdir /app \
     && chown -R ${XP_USER}:${XP_USER_GROUP} /app
 
-COPY --from=api-builder --chown=${XP_USER}:${XP_USER_GROUP} /app/api/experiments.yaml /app/
-COPY --from=api-builder --chown=${XP_USER}:${XP_USER_GROUP} /app/api/schema.yaml /app/
+COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/experiments.yaml /app/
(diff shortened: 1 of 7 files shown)


Dockerfile
```diff
@@ -21,6 +21,8 @@ RUN addgroup -S ${XP_USER_GROUP} \
 
 COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/* /app/
 COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/database /app/database/
+# read+write access for owner/group but no write access for others
+RUN chmod -R 775 /app/xp-management
```
krithika369, line 25: * Not sure if the chmod has some unintended impact. Are we still able to build and deploy the internal image on top of this? Can it be tested prior to merging this?
* Do we need a similar change for the treatment service too?

Dockerfile
```diff
@@ -19,8 +35,10 @@ RUN addgroup -S ${XP_USER_GROUP} \
     && mkdir /app \
     && chown -R ${XP_USER}:${XP_USER_GROUP} /app
 
-COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/* /app/
-COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/database /app/database/
+COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/experiments.yaml /app/
```
krithika369, line 38: Nit: Can we do `COPY --chown=${XP_USER}:${XP_USER_GROUP} management-service/bin/*.yaml /app/` instead of 2 separate commands to copy the OpenAPI specs?

Makefile
```diff
@@ -34,7 +34,7 @@ lint: lint-python lint-go
 .PHONY: vendor
 vendor:
 	@echo "Fetching dependencies..."
-	go mod vendor
+	cd management-service && go mod vendor
```
krithika369, line 37: We should probabl also add commands to vendor the `common` and `clients` modules before vendoring the management service.

management-service/appcontext/appcontext.go
```diff
@@ -30,36 +32,47 @@ func NewAppContext(db *gorm.DB, authorizer *mw.Authorizer, cfg *config.Config) (
...
 	projectSettingsSvc := services.NewProjectSettingsService(&allServices, db)
 
+	log.Println("Initializing segment history service...")
 	segmentHistorySvc := services.NewSegmentHistoryService(db)
+	log.Println("Initializing segment service...")
 	segmentSvc := services.NewSegmentService(&allServices, db)
 
+	log.Println("Initializing treatment history service...")
 	treatmentHistorySvc := services.NewTreatmentHistoryService(db)
+	log.Println("Initializing treatment service...")
```
krithika369, line 67: These seem like Debug logs. It would be enough to just have error logs if the initialization fails. Alternatively, if we want to keep these logs, we should log them at the appropriate level.

krithika369 (approved):
Left some small comments, the rest LGTM. Thanks @terryyylim !

PR #67, opened 2023-03-28
Update Google client initialisation to use MLP helper method

deadlycoconuts:
With the introduction of a helper method in MLP (https://github.com/gojek/mlp/pull/78) to abstract away the creation of a Google client to append ID tokens to the headers of outgoing requests, that is authenticated with service or user accounts, this PR imports the said helper method and uses it to initialise its Google client.

This ensures that all outgoing requests from the Turing API server contain ID tokens in their headers.

Fixes #


treatment-service/models/storage.go
```diff
@@ -571,23 +573,21 @@ func NewLocalStorage(
...
-			creds, err := google.CredentialsFromJSON(context.Background(), data, GoogleOAuthScope)
-			if err != nil {
-				return nil, err
-			}
-			googleClient = oauth2.NewClient(context.Background(), creds.TokenSource)
+			googleClient, err = idtoken.NewClient(
+				context.Background(),
+				defaultCaraMLAudience,
+				idtoken.WithCredentialsFile(filepath),
+			)
```
deadlycoconuts, line 585: Will be replacing this block with the new method `InitGoogleClientFromCredentialsFile` introduced in https://github.com/gojek/mlp/pull/79 (after this gets approved and merged) which initialises a Google client a specified file path for both service AND user accounts.

krithika369 (approved):
@deadlycoconuts, thanks for also testing out the plugin auth behavior and handling it. LGTM.

PR #69, opened 2023-04-13
Update MLP dependency references

terryyylim:

`mlp` dependency has been moved under `caraml-dev` organisation via https://github.com/caraml-dev/mlp and this PR updates XP to reflect that change.

Turing plugin's dependency is also updated and the `replace` directive for `gojek/mlp` can be removed due to the update to the MLP dependency.

`N/A`

**Does this PR introduce a user-facing change?**:

```release-note
NONE
```


plugins/turing/go.mod
```diff
@@ -4,35 +4,35 @@ go 1.18
 
 require (
 	bou.ke/monkey v1.0.2
+	github.com/caraml-dev/mlp v1.7.6
```
krithika369, line 7: Shall we use the same version of the MLP API as the management and treatment services, for the plugin too?

ui/package.json
```diff
@@ -4,10 +4,10 @@
   "private": true,
   "homepage": "/xp",
   "dependencies": {
+    "@caraml-dev/ui-lib": "^1.7.5-build.5-59f13e1",
```
krithika369, line 7: Can we update this to version 1.7.6 now?

plugins/turing/go.mod
```diff
@@ -125,12 +125,9 @@ require (
 )
 
 replace (
-	// To remove once the next version of Turing (v1.9.0) has been cut and released
-	github.com/caraml-dev/turing/engines/experiment => github.com/caraml-dev/turing/engines/experiment v0.0.0-20230104094236-477ac2392c59
-	github.com/caraml-dev/turing/engines/router => github.com/caraml-dev/turing/engines/router v0.0.0-20230104094236-477ac2392c59
+	github.com/caraml-dev/turing/engines/experiment => github.com/caraml-dev/turing/engines/experiment v0.0.0-20230414061300-682496141466
```
krithika369, line 128: Could you also mention the update to the Turing dependency in the PR description? This is needed purely for the update to the MLP dependency (which is why the replace directive for MLP below can be removed), right?

krithika369 (approved):
Left some small comments. LGTM. Thanks, @terryyylim !

plugins/turing/go.mod
terryyylim: Yes, that's my understanding as well!

plugins/turing/go.mod
terryyylim: Actually, I think it makes more sense to update management and treatment services to be `v1.7.6`, wdyt?

PR #70, opened 2023-04-13
Add gorm config to update nested fields

deadlycoconuts:
Similar to https://github.com/caraml-dev/turing/pull/331, the Gorm library has been updated to v2 in https://github.com/caraml-dev/xp/pull/43. As there was a bug discovered in how we were using the Gorm methods to save to the DB whereby nested fields do not get updated in their respective tables, we're now updating the Gorm configs in the XP repository too to ensure that this bug does not occur again (even though we are currently not saving nested fields in the DB in XP).

Fixes #


krithika369 (approved):
LGTM. Thanks, @deadlycoconuts !

PR #71, opened 2023-04-17
Revert PR (#70) - addition of gorm config to update nested fields in the db

deadlycoconuts:
This PR reverts the changes introduced in #71, as there is no need to enforce all associations of a Gorm object to be saved. A similar change was first introduced in https://github.com/caraml-dev/turing/pull/331 but it has since been removed in favour of a more precise manner of updating tables that contain nested fields of a Gorm object.

Fixes #


krithika369 (approved):
Thanks @deadlycoconuts !

PR #73, opened 2023-04-20
Update CICD and Go version

terryyylim:

This PR does 2 things:
1. Updates all modules to Go version 1.20
2. Skip linting and tests jobs when merged to `main` branch since these jobs are already triggered in the PR and not necessary to be repeated

`NONE`


treatment-service/models/storage_test.go
```diff
@@ -528,6 +528,9 @@ func TestDumpExperiments(t *testing.T) {
 			1: {NewExperimentIndex(e)},
 		},
 	}
+	// Need to hardcode this to testdata/experiments_dump.json ExperimentId
+	// because we are randomly generating Id in newTestXPExperiment call
+	e.Id = 81
```
krithika369, line 533: I see.. Can we instead make the function `newTestXPExperiment` accept a parameter to explicitly pass in the ID we want to use ?

.github/workflows/xp.yml
```diff
@@ -47,6 +48,7 @@ jobs:
         run: make lint-python
 
   lint-go:
+    if: github.head_ref != 'refs/heads/main'
```
krithika369, line 51: Doesn't this mean we will run this step for tags too? Should we split the commit to non-main branch from the release steps (release from `main` or from `tag`) into 2 files?

terryyylim: Are you proposing for `release-rules` and `release` steps to be in a separate file? And that all `lint` + `test` steps to be only for PRs?

krithika369: Not exactly.. Something like:
1. test.yaml - this one has all the lint / test steps for the "commit to non-main branch" event (cmiiw but with the current rule `if: github.head_ref != 'refs/heads/main'`, the lint / test will also be run for tag creation, which we want to avoid).
2. release.yaml - this one has the release steps (build & publish) for the main branch run / tag create / manual workflow dispatch events.

If we split it this way, we can have 2 separate files with the condition for triggering the workflow with the `on` tag rather than the `if` spec per task.

terryyylim: >the lint / test will also be run for tag creation, which we want to avoid

Oh then I guess I misunderstood, so CMIIW, but we want lint/test to run in non-main branch, so that we can test the changes in the event we cherry-pick commits prior to cutting a new tag right?

> release.yaml - this one has the release steps (build & publish) for the main branch run / tag create / manual workflow dispatch events.

So are you suggesting to do away with the current `workflow_call` trigger?

treatment-service/models/storage_test.go
krithika369: @terryyylim, I believe this comment is not addressed yet. Could you check?

krithika369: Hmm actually my bad - I see the function `newTestXPExperiment` is modified.

So why is this still needed?

Screenshot 2023-05-15 at 8 07 55 AM

krithika369 (approved):
Left on comment. The rest LGTM. Thanks, @terryyylim !

treatment-service/models/storage_test.go
terryyylim: Oh, which part would you be referring to? Passing in `81`?

krithika369: Yes

krithika369: Ahh I think I got it now - the 81 is now the parameter. Looks like the comment above that is outdated - it's no longer any "hardcoding", right?

terryyylim: Yes correct!

PR #74, opened 2023-05-11
Set env config to use _ delimiter instead of ::

shydefoo:
* This PR replaces the `::` delimiter with `_` so `DBCONFIG::PASSWORD` becomes `DBCONFIG_PASSWORD`
* This is to allow the setting of environment variables in `env` field in K8s PodSpec, which opens up the option to set env vars using ConfigMaps or Secrets
* Currently caraml-dev/helm-charts https://github.com/caraml-dev/helm-charts/actions/runs/4943497133/jobs/8838050166?pr=243 is failing because of the use of this delimiter

<!--
*Automatically closes linked issue when PR is merged.
Usage: `Fixes #<issue number>`, or `Fixes (paste link of issue)`.
nil

cc @mbruner

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ Instructions as described in README of https://github.com/caraml-dev/mlp.
 
 #### b. Starting XP
 
-Prior to starting XP, we'll need to ensure correct MLP API is correctly set in the config file, i.e management-service/config/example.yaml, and setting the correct `MLPConfig::URL` value.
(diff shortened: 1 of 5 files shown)


krithika369 (approved):
Thanks for this change, @shydefoo. I think the original motivation was to potentially cater to some configs that had `_` as a part of their name rather than use it as a delimiter. We don't have such cases today, so it's ok to use the standard `_`.

PR #75, opened 2023-05-17
Update mlp dependency

deadlycoconuts:
This PR updates the version of MLP uses in order to incorporate this change [here](https://github.com/caraml-dev/mlp/pull/84), which allows XP to use workload identity-enabled service account to authenticate all its outgoing resources.

Fixes #


krithika369 (approved):
LGTM. Thanks, @deadlycoconuts !

PR #76, opened 2023-06-14
feat: Add authz caching config

krithika369:

This PR largely replicates the changes for the API in https://github.com/caraml-dev/mlp/pull/87, to have the option of caching authz responses in the XP Management service. As a part of this implementation, the API dependency on MLP has been upgraded.


management-service/services/mlp_service.go
```diff
@@ -97,7 +97,7 @@ func (service mlpService) refreshProjects() error {
 	ctx, cancel := context.WithTimeout(context.Background(), mlpQueryTimeoutSeconds*time.Second)
 	defer cancel()
 
-	projects, resp, err := service.mlpClient.api.ProjectApi.ProjectsGet(ctx, nil)
+	projects, resp, err := service.mlpClient.api.ProjectApi.V1ProjectsGet(ctx, nil)
```
krithika369, line 100: This is a consequence of removing `v1/` from the base path in: https://github.com/caraml-dev/mlp/pull/85

The change was applied to Turing / Merlin already. Eg: https://github.com/caraml-dev/turing/pull/338

deadlycoconuts (approved):
LGTM! 🚀 Thanks for the changes!

deadlycoconuts:
Just curious, was it intentional to no longer create branches on one's fork for PRs? 😅

krithika369:
> Just curious, was it intentional to no longer create branches on one's fork for PRs? 😅

Since this isn't the practice in Merlin and MLP repos, I decided to just stick to branches for everything (in my new laptop). But I realised that for Turing and XP, the fork's CI steps are cleaner. Will get back to using them for the newer PRs. :)

PR #78, opened 2024-04-24
doc: Update doc on switchback experiment

krithika369:

As of now, randomized switchback experiments do not support varying the assigned treatment by some value (see [here](https://github.com/caraml-dev/xp/blob/v0.13.0/treatment-service/services/treatment_service.go#L69) - we set `""` as the randomization value and therefore the [seed](https://github.com/caraml-dev/xp/blob/v0.13.0/treatment-service/services/treatment_service.go#L172) is not influenced by it). So, all requests matching a switchback experiment in a given time interval will be assigned the same treatment. Updating the doc to reflect this.


docs/concepts.md
```diff
@@ -56,7 +56,7 @@ XP support various experiment types.
...
-- **Randomized Switchback Experiments** - This is a hybrid between the A/B experiments and Switchbacks. These experiments are Switchbacks by nature (they have a time interval). In addition, they can have a traffic allocation on each of the treatments. Thus, at every new interval, the selection of the treatment is not cyclical, but randomized. In the default mode, all requests in a given time interval will still receive the same treatment. However, it is also possible to vary this.
+- **Randomized Switchback Experiments** - This is a hybrid between the A/B experiments and Switchbacks. These experiments are Switchbacks by nature (they have a time interval). In addition, they can have a traffic allocation on each of the treatments. Thus, at every new interval, the selection of the treatment is not cyclical, but randomized. All requests in a given time interval will receive the same treatment.
```
deadlycoconuts, line 59: Just to clarify what I understand from these changes, this is what the doc is trying to clarify right?

(Old) potentially incorrect interpretation: The treatment received by a request in a specific interval is dependent on the selected experiment, the time interval during which the request is received, and some other configurable value(?) from the request.

(New) correct interpretation: The treatment received by a request in a specific interval is purely determined by the selected experiment and the time interval during which the request is received.

krithika369: That is correct. The "some other configurable value(?) from the request" was never implemented.

tests/e2e/test_validate_treatment.py
```diff
@@ -1,16 +1,18 @@
 from xp_client import XPClient
 
+def test_validate_valid_treatment_with_url(xp_client: XPClient, httpserver):
+    # Mock external validation endpoint
+    httpserver.expect_request("/external", method="POST").respond_with_json({})
 
-def test_validate_valid_treatment_with_url(xp_client: XPClient):
     data = {
-        "validation_url": "http://www.mocky.io/v2/5e4cacd4310000e1cad8c073",
+        "validation_url": httpserver.url_for("/external"),
```
krithika369, line 8: The old mocky URL had expired so the e2e failed. Replacing that with a local test httpserver.

docs/concepts.md
deadlycoconuts: Thanks for the clarification!

treatment-service/go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/caraml-dev/mlp v1.7.7-0.20230428104022-779530aec912
 	github.com/caraml-dev/xp/clients v0.0.0-00010101000000-000000000000
 	github.com/caraml-dev/xp/common v0.0.0
-	github.com/confluentinc/confluent-kafka-go v1.8.2
+	github.com/confluentinc/confluent-kafka-go v1.9.2
```
krithika369, line 11: The older kafka library version doesn't build on M1. Upgrading it to the latest.

leonlnj (approved):

PR #80, opened 2024-07-23
fix(ui): Fix yup validation bug

deadlycoconuts:
Similar to https://github.com/caraml-dev/turing/pull/387, this PR to fixes broken validation schemas caused by the upgrade of the `yup` package in PR #591 - these changes can largely be grouped into the following:

- When using the same `Schema.when` function but passing a function directly as an argument instead of the builder object, the signature of the expected function has also been updated from `(value, schema)=> Schema): Schema` to `(values: any[], schema) => Schema): Schema`.

  Existing schemas that do not follow this convention now have been updated to reflect the new expected array:

  Example:
  ```javascript
  // before
  some_field: yup.string().when("some_other_field", (some_other_field, schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),

  // after
  some_field: yup.string().when("some_other_field", ([some_other_field], schema) =>
    some_other_field ? yup.string().required("this is needed") : schema
  ),
  ```
- The `mixed` schema has been severely nerfed and is now no longer the base class for all the other schemas, and as such no longer supports a lot of the other functions that we were previously calling. Schemas that have fields which use the `mixed` schema have been updated to use a schema (e.g. `string`) corresponding to its expected primitive data type wherever possible (where we know that only a specific type is expected).

This PR also bumps up the version of the `caraml-dev/ui-lib` library to the latest version and removes custom css styles for the experiments search sidebar which no longer work.

Fixes #

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -8,7 +8,7 @@
   "//": "[@sentry/browser] pinned to 7.118.0 because craco/module federation has issues resolving this dependency; see",
   "//": "https://github.com/caraml-dev/turing/pull/384#discussion_r1666418144 for more details",
   "dependencies": {
-    "@caraml-dev/ui-lib": "^1.13.0-build.3-a234b6b",
+    "@caraml-dev/ui-lib": "^1.13.0-build.4-09c363a",
     "@elastic/datemath": "^5.0.3",
     "@elastic/eui": "88.2.0",
     "@emotion/css": "^11.11.2",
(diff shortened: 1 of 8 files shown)


leonlnj (approved):
`The mixed schema has been severely nerfed` lol. 

Thanks for the mammoth upgrades!

deadlycoconuts:
Ahaha thanks for the quick approval! Merging this now! 🚀

PR #82, opened 2024-08-05
chore(tests): Update test files to work with Docker Compose v2 

deadlycoconuts:
It seems like v1 of Docker Compose has just been [removed from all ubuntu GitHub runners](https://github.com/actions/runner-images/issues/9692) as of the 29th of July (and has been ongoing since that day). This has broken the CICD pipelines of lots of repos (just take a look at the number of times that issue has been referenced), including XP. As such we can no longer run commands like `docker-compose` in the GitHub runners anymore and need to start updating them to use the v2 commands (to `docker compose` for this particular example).

For XP in particular, what broke were the Treatment Service integration tests, where we were using the `testcontainers` package to spin up Kafka. This version however, is also similarly super old, and calls `docker-compose` (v1) to spin containers in Go code. As such, this PR updates the version of `testcontainers` (which also requires some additional changes to the methods we need to call) so that the Docker Compose v2 commands are used instead.

Summary of Changes:
- Renaming of `docker-compose` to `docker compose`
- Bumping of `testcontainers` version
- Bumping of Go version (I didn't want this but we need to bump to the later/latest version of `testcontainers`)
- Bumping of the Go linter versions (in order to work with the newer Go version)
- Bumping of all other random dependencies after running `go mod tidy` in all places
- Replacement of the deprecated `io/ioutil` package methods with `io` methods (this mostly applied to the autogenerated code in `clients/management/managementclient.go` and `clients/treatment/treatmentclient.go`; I could've regenerated the code with a newer version of the code generator but I just decided to use my IDE to replace them)
- Refactoring of the integration test setup script in `treatment-service/integration-test/fetch_treatment_it_test.go` to use the new Docker Compose v2 commands
- Change to the existing unit test in `treatment-service/models/storage_test.go` to ensure that an existing unit test no longer breaks because it's performing a check on a randomly generated ID

Fixes #


treatment-service/models/storage_test.go
```diff
@@ -523,6 +523,9 @@ func TestDumpExperiments(t *testing.T) {
 		time.Date(2022, 1, 2, 3, 5, 7, 0, time.UTC),
 	), segmentersType)
 	require.NoError(t, err)
+
+	// Manually set experiment id so it matches the id in the dump
+	e.Id = 81
```
deadlycoconuts, line 528: This id is now set manually because for some reason the unit tests are failing as this ID field (it's generated randomly [here](https://github.com/caraml-dev/xp/blob/6ef47af4e979f7033133c40f3eff281a7ae85114/treatment-service/models/storage_test.go#L54)) is now being compared with a hardcoded id value found [here](https://github.com/caraml-dev/xp/blob/6ef47af4e979f7033133c40f3eff281a7ae85114/treatment-service/testdata/experiments_dump.json#L20). This might be happening because of the upgrade to the proto version I believe.

treatment-service/integration-test/fetch_treatment_it_test.go
```diff
@@ -370,11 +371,11 @@ func (suite *TreatmentServiceTestSuite) SetupSuite() {
...
-	kafka := testcontainers.NewLocalDockerCompose(composeFilePaths, "kafka")
-	execError := kafka.
-		WithCommand([]string{"up", "-d"}).
-		Invoke()
-	err = execError.Error
+	kafka, err := compose.NewDockerComposeWith(compose.WithStackFiles(composeFilePaths...), compose.StackIdentifier("kafka"))
+	if err != nil {
+		panic(err)
+	}
+	err = kafka.Up(ctx, compose.Wait(true))
```
deadlycoconuts, line 378: These is the new way we're supposed to run Docker compose (`testcontainers` uses the v2 commands with `compose.NewDockerComposeWith`). I know I mentioned that I took a long time to debug the connectivity issues with the spun up Kafka instance; in the end I realised that the connectivity was working fine all along -.-" I got confused by some logs in the CICD pipeline which made me think that there was an issue with the networking in Docker.

treatment-service/integration-test/docker-compose/kafka/docker-compose.yaml
```diff
@@ -1,5 +1,3 @@
-version: '3'
-
```
deadlycoconuts, line 2: In Docker Compose v2, this version field is no longer used: https://stackoverflow.com/a/76157215

management-service/go.mod
```diff
@@ -1,10 +1,12 @@
 module github.com/caraml-dev/xp/management-service
 
-go 1.18
+go 1.21
+
+toolchain go1.22.5
```
tiopramayudi, line 5: what is the purpose of this? And why the version if difference compare to the above one?

treatment-service/models/storage_test.go
tiopramayudi: How was this even working previously when the id is generated randomly but the expected id is set to 81 ?

tiopramayudi (commented):
Add some comments, the rest is LGTM! Thanks @deadlycoconuts

management-service/go.mod
deadlycoconuts: That's autogenerated after running `go mod tidy` when I've just only updated the `testcontainers-go` package. I would assume that ideally the go version would've been updated to `1.22` and above also, but I'm guessing that because there are some (other) dependencies which cannot be upgraded past version `1.21`, the end result was was this toolchain being specified at a higher version than the actual go version itself. 

I'm not super clear on how these toolchain stuff work, but there's this (very long article) by Go when they first released toolchains as part of bundled Go distribution [here](https://go.dev/doc/toolchain). In particular, there are these paragraphs that kind of suggest that having a higher toolchain version is like a compromise to maintain backward compatibility of some packages:

> In the standard configuration, the go command uses its own bundled toolchain when that toolchain is at least as new as the go or toolchain lines in the main module or workspace. For example, when using the go command bundled with Go 1.21.3 in a main module that says go 1.21.0, the go command uses Go 1.21.3. When the go or toolchain line is newer than the bundled toolchain, the go command runs the newer toolchain instead. For example, when using the go command bundled with Go 1.21.3 in a main module that says go 1.21.9, the go command finds and runs Go 1.21.9 instead. It first looks in the PATH for a program named go1.21.9 and otherwise downloads and caches a copy of the Go 1.21.9 toolchain. This automatic toolchain switching can be disabled, but in that case, for more precise forwards compatibility, the go command will refuse to run in a main module or workspace in which the go line requires a newer version of Go. That is, the go line sets the minimum required Go version necessary to use a module or workspace.

> Modules that are dependencies of other modules may need to set a minimum Go version requirement lower than the preferred toolchain to use when working in that module directly. In this case, the toolchain line in go.mod or go.work sets a preferred toolchain that takes precedence over the go line when the go command is deciding which toolchain to use.

treatment-service/models/storage_test.go
deadlycoconuts: I have no idea to be honest 😂 There's actually another field in the same struct (`name`) that's being generated with a new UUID each time https://github.com/caraml-dev/xp/blob/6ef47af4e979f7033133c40f3eff281a7ae85114/treatment-service/models/storage_test.go#L47 and somehow this doesn't fail the deepequal comparison. I don't think we changed anything else so it must've been one of the upgrades to the protobuf/reflect packages that changed something 🤔

tiopramayudi (approved):

PR #84, opened 2024-08-08
chore(docker): Update base docker images

deadlycoconuts:
With the upgrade of Go from 1.18 to 1.21 in all of XP's components, we need to update the docker base image used to build them. This PR updates the version of Go in these base images from 1.18 to 1.21.

Fixes #


deadlycoconuts:
Thanks a lot for the quick review!

PR #85, opened 2024-08-08
fix(clients): Update clients go sum file

deadlycoconuts:
It seems like the management service build was [failing](https://github.com/caraml-dev/xp/actions/runs/10298395475/job/28503879095) because one of the go sum files wasn't updated in `clients`. This PR updates that go sum file.

Fixes #


tiopramayudi (approved):

deadlycoconuts:
Thanks for the quick review once again!

PR #86, opened 2024-08-12
fix(ui): Add helper function to cast string to array before yup validation

deadlycoconuts:
It seems like after updating yup/elastic UI packages, array values in the `EuiTextArea` are expressed as a JSON literal, e.g. `"[\"new\"]"` and these values no longer get automatically parsed into a JSON array at the yup validation stage. As a result, all the `.array()` validations that were performed no longer work and fail (because they report a string being present instead). The screenshot attached shows how the validation reports an error even though nothing is wrong with the input in the text fields.

An additional step is thus introduced to manually parse the literal input into a JSON array (failing the validation check if it isn't), and to validate the parsed array with respect to the original schema. 

Screenshot 2024-08-12 at 13 41 11

Fixes #

### DIFF
--- a/ui/src/settings/segmenters/components/form/validation/schema.js
+++ b/ui/src/settings/segmenters/components/form/validation/schema.js
@@ -55,30 +55,28 @@ const optionsSchema = yup
     }
   );
 
-const validateArrayString = (arraySchema, arrayName) => {
-  return yup
-    .string()
-    .test(
-      `${arrayName} valid JSON array`,
-      `${arrayName} must be a valid JSON array`,
-      (array) => {
-        if (array !== "") {
-          try {
-            var parsedArray = JSON.parse(array);
-            if (typeof parsedArray != "array" && !Array.isArray(parsedArray)) {
-              return false;
-            }
-            return arraySchema
-              .validateSync(parsedArray);
-          } catch (e) {
+const validateArrayString = (arraySchema, arrayName) => yup
+  .string()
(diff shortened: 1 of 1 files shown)


ui/src/settings/segmenters/components/form/validation/schema.js
```diff
@@ -55,13 +55,35 @@ const optionsSchema = yup
...
+const validateArrayString = (arraySchema, arrayName) => yup
+  .string()
+  .test(
+    `${arrayName} valid JSON array`,
+    `${arrayName} must be a valid JSON array`,
+    (array) => {
+      if (array !== "") {
+        try {
+          var parsedArray = JSON.parse(array);
+          if (typeof parsedArray != "array" && !Array.isArray(parsedArray)) {
```
bthari, line 67: sorry for my limited knowledge in javascript, but isn't `typeof (array)` is `object`?

deadlycoconuts: Yes you're right `typeof someArray` always returns an object. I modified this check incorrectly 😅. I've changed it to `typeof parsedArray != "object" || !Array.isArray(parsedArray)` now, which is similar to what we already have for some of the other tests in this file.

bthari (approved):
Thank you for this fix! LGTM 🚀

deadlycoconuts:
Thanks for the quick review! I'll merge this soon!

PR #87, opened 2024-12-24
Treatment service Poller

rautelaanirudh:
The intention of this PR is to provide the option of utilising a poller in the treatment service instead of pub/sub. The poller will be responsible for pulling the changes periodically instead of subscribing to the management service topic.

Poller can be enabled from treatment service config

```
PollerConfig:
  Enabled: true
  PollInterval: 10s
```

Treatment service can be used in cloud environments where pub/sub is currently not supported e.g. Tencent cloud.


treatment-service/appcontext/appcontext.go
```diff
@@ -92,18 +93,17 @@ func NewAppContext(cfg *config.Config) (*AppContext, error) {
 		return nil, err
 	}
 
-	log.Println("Initializing message queue subscriber...")
 	var messageQueueService messagequeue.MessageQueueService
-	switch cfg.MessageQueueConfig.Kind {
-	case common_mq_config.NoopMQ:
+	if cfg.PollerConfig.Enabled || cfg.MessageQueueConfig.Kind == common_mq_config.NoopMQ {
```
deadlycoconuts, line 97: I think we can just leave the polling config and message queue config as independent configuration i.e. whether the poller is set up or not doesn't have an impact on whether the message queue is configured or not. In other words, we can have the poller enabled BUT with a valid message queue (pubsub in this case) enabled also.

That is to say, let's remove the checks on the poller config here.

treatment-service/server/poller.go
```diff
@@ -0,0 +1,72 @@
...
+		pollerConfig: pollerConfig,
+		localStorage: localStorage,
+		control: PollerControl{
+			stopChannel: make(chan struct{}),
+		},
+	}
+}
+
+func (p *Poller) Start() {
+	p.control.startOnce.Do(func() {
```
deadlycoconuts, line 39: Why do we need this?

treatment-service/server/poller.go
```diff
@@ -0,0 +1,72 @@
...
+					ticker.Stop()
+					return
+				}
+			}
+		}()
+	})
+}
+
+func (p *Poller) Stop() {
+	p.control.stopOnce.Do(func() {
```
deadlycoconuts, line 63: Similar question here, why do we need to use a `sync.Once` object here?

treatment-service/server/poller.go
```diff
@@ -0,0 +1,72 @@
...
+		control: PollerControl{
+			stopChannel: make(chan struct{}),
+		},
+	}
+}
+
+func (p *Poller) Start() {
+	p.control.startOnce.Do(func() {
+		ticker := time.NewTicker(p.pollerConfig.PollInterval)
+		p.control.waitGroup.Add(1)
```
deadlycoconuts, line 41: Hmm what's the point of adding this wait group? There isn't anything to complete that we're waiting for before `p.control.waitGroup.Done()` gets called right?

deadlycoconuts (changes requested):
Thanks a lot of this PR! There are just a couple of small comments I left, especially on the polling logic which I didn't really understand why it needed to be implemented as such.

treatment-service/appcontext/appcontext.go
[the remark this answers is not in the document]
rautelaanirudh: If the poller config is enabled, we initialize a queue of kind `NoopMQ` instead of `PubSubMQ` which also means that it is not a valid message queue. Any specific reason why we should have a queue if we actually enable poller since the poller would not utilise any queue ?

treatment-service/config/config.go
```diff
@@ -34,6 +34,7 @@ type Config struct {
 	MonitoringConfig        Monitoring                          `json:"monitoring_config"`
 	SwaggerConfig           SwaggerConfig                       `json:"swagger_config" validate:"required,dive"`
 	SegmenterConfig         map[string]interface{}              `json:"segmenter_config"`
+	PollerConfig            PollerConfig                        `json:"poller_config" validate:"required,dive"`
```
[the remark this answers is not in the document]
rautelaanirudh: Thanks for the suggestion. Fixed

treatment-service/appcontext/appcontext.go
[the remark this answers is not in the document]
deadlycoconuts: > Any specific reason why we should have a queue if we actually enable poller since the poller would not utilise any queue ?

I don't mean that we must initialise a `PubSubMQ` if the poller is enabled, but rather, I was saying that we can allow the `PubSubMQ` to be initialised (or not) independently of whether the poller is enabled or not. 

Currently, there are 2 separate set of configs (`PollerConfig` and `MessageQueueConfig`) which seem to be interdependent (flipping a flag on one config changes the behaviour of the other), which can be confusing for someone having to manage configs. It's definitely okay to the keep the checks that you've implemented but I would've instead then suggested refactoring both `PollerConfig` and `MessageQueueConfig` into a single struct (with some renaming and additional config validity checks) so it becomes more immediately obvious that they both affect the same functionality (i.e. the way the Treatment Service gets updated configuration from the Management Service).

So going back to your original question, nope there's no specific reason why someone would need to have both enabled, but I was only suggesting that change to keep the current changes simple.

treatment-service/server/poller.go
rautelaanirudh: Actually I just realised since we are only invoking the poller start method from a single place, we don't need it. Thanks for pointing that out.

treatment-service/server/poller.go
rautelaanirudh: Yes, removed it now.

treatment-service/appcontext/appcontext.go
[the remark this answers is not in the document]
rautelaanirudh: Ack, it makes sense to decouple the `PubSubMQ` initialisation then.

treatment-service/server/poller.go
deadlycoconuts: Thanks!

deadlycoconuts (approved):
Thanks a lot for all the changes! I've just one last tiny comment but feel free to merge this once that's addressed! :D 🚀

treatment-service/appcontext/appcontext.go
```diff
@@ -92,7 +93,6 @@ func NewAppContext(cfg *config.Config) (*AppContext, error) {
 		return nil, err
 	}
 
-	log.Println("Initializing message queue subscriber...")
```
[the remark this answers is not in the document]
rautelaanirudh: Sorry, it got removed unintentionally. Thanks for pointing that out :-)

PR #90, opened 2025-02-28
feat: improve locking: lock only after fetch completed

faqiharifian:

When pubsub disabled and poll enabled, every time poller runs XP treatment service tries to acquire write lock and the treatment service refreshes the local storage by doing multiple calls to management service and update the data/maps.

This causes other part that use RLock to be blocked until all the calls and update are done. 

This MR moved the write lock to after all the calls are completed, this reduces duration of RLock being blocked

### DIFF
--- a/treatment-service/models/storage.go
+++ b/treatment-service/models/storage.go
@@ -254,9 +254,6 @@ func (i *ExperimentIndex) checkSegmentHasWeakMatch(segmentName string) bool {
 }
 
 func (s *LocalStorage) InsertProjectSettings(projectSettings *pubsub.ProjectSettings) error {
-	s.Lock()
-	defer s.Unlock()
-
 	// check that settings with the same Id doesn't exist
 	for _, existingSettings := range s.ProjectSettings {
 		if existingSettings.GetProjectId() == projectSettings.GetProjectId() {
@@ -264,7 +261,8 @@ func (s *LocalStorage) InsertProjectSettings(projectSettings *pubsub.ProjectSett
 		}
 	}
 
-	s.ProjectSettings = append(s.ProjectSettings, projectSettings)
+	s.swapProjectSettings(append(s.ProjectSettings, projectSettings))
+
 	// Update project segmenters on creation
 	if err := s.initProjectSegmenters([]*pubsub.ProjectSettings{projectSettings}); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


mbruner (approved):

treatment-service/models/storage.go
```diff
@@ -254,21 +254,22 @@ func (i *ExperimentIndex) checkSegmentHasWeakMatch(segmentName string) bool {
...
 func (s *LocalStorage) InsertProjectSettings(projectSettings *pubsub.ProjectSettings) error {
-	s.Lock()
-	defer s.Unlock()
-
 	// check that settings with the same Id doesn't exist
-	for _, existingSettings := range s.ProjectSettings {
-		if existingSettings.GetProjectId() == projectSettings.GetProjectId() {
-			return nil
-		}
+	existingProjectSettings := s.findProjectSettingsById(ProjectId(projectSettings.GetProjectId()))
```
bthari, line 258: if the locking is not running in here, is it possible that there will be two client who requests the same `GetProjectId`, and therefore the `s.ProjectSettings` in [here](https://github.com/faqiharifian/xp/blob/cdb488b2c0fcc64d4a3ad825bec516d69bd99f36/treatment-service/models/storage.go#L271-L272) will have duplicate item?

treatment-service/models/storage.go
```diff
@@ -284,30 +285,55 @@ func (s *LocalStorage) UpdateProjectSettings(updatedProjectSettings *pubsub.Proj
...
 
 func (s *LocalStorage) FindProjectSettingsWithId(projectId ProjectId) *pubsub.ProjectSettings {
+	projectSettings := s.findSubscribedProjectSettingsById(projectId)
+	if projectSettings != nil {
+		return projectSettings
+	}
+
+	// In case new project was just created and we are subscribed to its ID
+	// we'll try to retrieve it from management service
+	projectSettings, err := s.fetchProjectSettingsWithId(projectId)
```
bthari, line 295: shouldn't we put the `!ContainsProjectId(s.subscribedProjectIds, projectId)` before this function? It is possible that `projectSettings` above is still nil, not because just the project can't be find, but also we don't subscribe to this project. If this is intended with the new change, perhaps can modify the comment on above this

treatment-service/models/storage.go
faqiharifian: I just moving things around to optimize the locking, there's no new logic added in this MR and this method is not used by the Poller changes we introduced previously. Will not do it in this MR.

treatment-service/models/storage.go
faqiharifian: Not sure about the intention of this, I just moving things around to optimize the locking, including the comment

PR #91, opened 2025-03-26
fix(treatment-service): Add pubsub message ordering flag

deadlycoconuts:
Message ordering is not turned on for PubSub subscriptions created by the Treatment Service (and by extension the XP plugin), which means that messages received are not guaranteed to be in order. This poses a problem when there is a high volume of messages sent by the Management Service in a short span of time, as the order of messages received is important since these messages directly impact how the correct treatment is assigned to each request (e.g. an experiment is repeatedly turned on and off over a short period of time). 

This PR thus turns on message ordering all subscriptions created to ensure that all instances of the Treatment Service/plugin   are consistent in serving the same treatment to each request by maintaining strict message ordering.

See for more details: https://cloud.google.com/pubsub/docs/ordering#go

Fixes #

### DIFF
--- a/common/testutils/pubsub.go
+++ b/common/testutils/pubsub.go
@@ -82,7 +82,7 @@ func CreateSubscriptions(client *pubsub.Client, ctx context.Context, topicNames 
 		topic := client.Topic(topicName)
 		subscription, err := client.CreateSubscription(ctx,
 			subscriptionId,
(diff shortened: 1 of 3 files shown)


bthari (approved):
woah, I just saw this MR again, but LGTM! Thanks for catching this!

deadlycoconuts:
Whoa thanks for the approval! I almost forgot about this! 🙏🏼

PR #93, opened 2026-08-14
fix(treatment-service): remove nested RLock deadlock and add optional LocalStorage call metrics

anantadwi13:

  1. **Fixes a deadlock** in `LocalStorage.findSubscribedProjectSettingsById`: it took `s.RLock()`
     and then, while still holding it, called `findProjectSettingsById`, which took `s.RLock()`
     again on the same goroutine. Go's `sync.RWMutex` blocks new readers once a writer is queued
     (to prevent writer starvation), so if a writer (`PollerService.Refresh` -> `Init()`, or a
     pubsub `InsertExperiment`/`UpdateProjectSettings` handler) queued between the outer and
     nested `RLock`, the nested call would block forever waiting for the writer, and the writer
     would block forever waiting for the outer `RLock` to release — a genuine circular-wait
     deadlock, suspected to be the root cause of a reported `GetTreatmentForRequest` hang.

     Fix: split the lock-acquiring `findProjectSettingsById` from a new lock-free
     `findProjectSettingsByIdLocked`, and have the already-locked caller invoke the latter
     directly instead of re-acquiring the lock.

  2. **Adds feature-flagged instrumentation** on `LocalStorage`'s public methods (a call counter
     and call-duration histogram, labeled by method name), so a stuck method — e.g. a lock
     deadlock like the one above — shows up as its call count stalling and duration climbing,
     instead of silently hanging with no observability.

     - New config flag: `Monitoring.LocalStorageMetricsEnabled` (default `false`).
     - Wired via a `SetMetricsRecorder` seam (`LocalStorageMetricsRecorder` interface) set from
       `appcontext.NewAppContext`, avoiding an import cycle between `models` and `services`.
     - New Prometheus metrics: `local_storage_calls_total`, `local_storage_call_duration_ms`.
     - Registered as Turing plugin collector metrics in `plugins/turing/runner/experiment_runner.go`.

     Tests: `storage_deadlock_test.go` deterministically reproduces the nested-RLock deadlock in
     isolation and stress-tests the real call path to confirm it no longer hangs;
     `storage_metrics_test.go` / `metric_service_test.go` cover the metrics recorder contract on
     both read and write paths.

### DIFF
--- a/treatment-service/models/storage.go
+++ b/treatment-service/models/storage.go
@@ -307,13 +307,21 @@ func (s *LocalStorage) findSubscribedProjectSettingsById(projectId ProjectId) *p
 		return nil
 	}
 
-	return s.findProjectSettingsById(projectId)
+	return s.findProjectSettingsByIdLocked(projectId)
 }
 
 func (s *LocalStorage) findProjectSettingsById(projectId ProjectId) *pubsub.ProjectSettings {
 	s.RLock()
 	defer s.RUnlock()
 
+	return s.findProjectSettingsByIdLocked(projectId)
+}
+
+// findProjectSettingsByIdLocked assumes the caller already holds s's read (or write) lock.
+// It must not itself call RLock/Lock, since sync.RWMutex blocks new readers once a writer is
+// queued -- a nested RLock on the same goroutine that already holds the lock would deadlock
+// against that queued writer.
+func (s *LocalStorage) findProjectSettingsByIdLocked(projectId ProjectId) *pubsub.ProjectSettings {
 	for _, settings := range s.ProjectSettings {
 		if ProjectId(settings.ProjectId) == projectId {
 			return settings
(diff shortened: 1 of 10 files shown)


treatment-service/models/storage.go
```diff
@@ -59,6 +59,28 @@ type LocalStorage struct {
...
+	// can't reach back into an instrumented LocalStorage method (directly or transitively) --
+	// that would turn this into unbounded recursion, not just a slow call.
+	metricsRecorder LocalStorageMetricsRecorder
+}
+
+// SetMetricsRecorder wires recorder into LocalStorage's public methods, which will start
+// reporting a call counter and call duration histogram to it. Call once, after construction;
+// leaving it unset (the default) means no metrics are reported.
+func (s *LocalStorage) SetMetricsRecorder(recorder LocalStorageMetricsRecorder) {
+	s.metricsRecorder = recorder
```
anantadwi13: possible race condition

